PR #24, opened 2023-02-24
refactor: Refactored the project structure in order to separate pages

frol:
Now, all the pages are defined in `src/gigs-board/pages/` folder, and they should use the common layout by rendering the `src/gigs-board/components/layout/Page` widget. Here is how it is done with the helper:

```
return widget("components.layout.Page", { children: pageContent });
```

Tip: In order to update the shared code from common.jsx file, run `npm install && npm run build`, which will update all the relevant blocks in all the source files.

I tested it on:
* testnet: https://test.near.social/#/gigs.frol14.testnet/widget/gigs-board.pages.Feed
* mainnet: https://near.social/#/frol.near/widget/gigs-board.pages.Feed?nearDevGovGigsContractAccountId=devgovgigs.near (notice that you can configure which contract to use through the props, while by default it falls back to the `widgetSrc`, meaning that it would try to use the account that has the widgets deployed)


frol:
The new widgets were deployed successfully: https://near.social/#/devgovgigs.near/widget/gigs-board.pages.Feed

I will replace old widgets with empty files, though I will keep Ideas and Post widgets to be proxies to the relevant gigs-board.pages

PR #42, opened 2023-03-03
show accounts that liked a post in a dropdown

### LINKED ISSUE #22: feat: Show accounts that liked a post in a dropdown
**Current State:** The Gigs Board does not display who liked a post:
Like-Posts

**Desired State:** Apply the same behavior from NEAR Social (showing users who liked the post + dropdown):
Like-Dropdown

ailisp:
This PR will close #22 

Preview:
image

image

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,7 @@
       "name": "devgigsboard-widgets",
       "version": "1.0.0",
       "license": "ISC",
-      "devDependencies": {
+      "dependencies": {
         "prettier": "^2.8.4",
         "replace-in-files": "^3.0.0"
       }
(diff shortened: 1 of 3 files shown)


frol:
@ailisp Let's replicate the UI instead of mixing the two:

Desired state:
Like-Dropdown

ailisp:
@frol Does red heart play well with blue "Reply" & "Expand Replies"? I feel my version looks more consistent. @ori-near what is your opinion?

frol:
We may keep the heart blue, but "Like (100)" duplicates the avatars and "and 97 others", and the hover blue background makes it impossible to read the text:

image

ailisp:
Current preview:

A post without likes:

image

A post with 1 like:
image

A post with 3 likes:
image

A post with more than 3 likes:
image

image

ori-near:
Hi @frol and @ailisp – thanks for the discussion above. I think @ailisp's latest solution is a good solution for now. And I think we can revisit the general UI / branding in the near future.

frol (approved):

Issue #46, opened 2023-03-06
feat: Improve the usability of moving posts through Kanban Boards


frol:
**Is your feature request related to a problem? Please describe.**

We want to make the funding process transparent, and currently, we use Kanban Board to represent the state of the sponsorship request. The process has multiple stages, and currently, the funding team has to learn how to edit posts to remove one label and add a new label to move a post from one Kanban Board column to another.

**Describe the solution you'd like**

There is no clear solution yet. We will revisit it once we collect more requirements.

**Describe alternatives you've considered**

* A simple button "Move to X" that will update the labels accordingly in a single click. See #47 for the details and the reasons we did not go further in that direction.
* We discussed with @ori-near that currently there are too many columns in the Funding board, and it would be great to hide the unnecessary complexity through the introduction of special sponsorship fields that would represent the timeline as a checklist while we will keep only 4 high-level funding states: New Funding Request, Application, Processing, Funded. With this change, we will need to design more like an end-to-end solution to the checklist and status updates.
* We previously talked about drag-n-drop UI, but it is not that easy to implement with the current NEAR Social.

**Additional context**

Permissions should be taken into account, so we don't confuse Gigs Board users who don't have permission to update the labels with this button (see #10).

ailisp:
@ori-near is this ticket still relevant? I don't see any kanban board now

PR #47, opened 2023-03-06
feat: Draft implementation of "Move to X" button for the posts on the Kanban Board

frol:
This is a draft implementation of the low-hanging fruit solution described in #46.

image

Check it on testnet: https://test.near.social/#/gigs.frol14.testnet/widget/gigs-board.pages.Boards?selectedBoardId=nearsocial

TODO:
* [ ] Do not display the button to those users who don't have permission to edit the labels
* [ ] Improve the design of the button
* [ ] Consider if there should be a button to move back (e.g. if it was moved by mistake)
* [ ] Consider making a special prop flag for a board where this "Move to" is applicable, as it does not make sense to move a post from a "widget" to "integration".
* [ ] Notify the involved parties about the update - it is not a blocker, but worth to think about

### DIFF
--- a/src/gigs-board/components/boards/KanbanBoard.jsx
+++ b/src/gigs-board/components/boards/KanbanBoard.jsx
@@ -85,6 +85,34 @@ const postsPerLabel = columnLabels.map((cl) => {
   }
 });
 
+function updatePostLabels(postId, oldLabel, newLabel) {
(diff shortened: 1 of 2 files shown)


frol:
After syncing with Ori, we decided that this implementation will only make a marginal improvement as it is common to also edit the description of the post to provide extra details (e.g. link to a transaction). Thus, I will not pursue this direction.

PR #85, opened 2023-04-07
Integrate fuzzy search 

### LINKED ISSUE #66: [Feed] Introduce search capability for posts
**Problem**
Currently, the Gigs Board only has a search by label function. Users are unable to search for posts using keyword, which makes it difficult to find relevant content.

**User Story**
As a Gigs Board user, I want to search for posts that mention a specific topic or keyword even if I don't know the exact phrasing or spelling. 

**Acceptance Criteria**
- Display a search input at the top of each page
- The search function uses fuzzy matching to return posts that match the search term, including partial matches and mispellings
- The search function is implemented using in-browser JavaScript to iterate through the post and do fuzzing matching instead of using middleware
- A UI element (e.g. spinning wheel) displays while the search is processing 
- The search function displays results as they are found, allowing users to see posts matching the search in real-time
- The search results should load quickly

**Note**
- This idea was featured as a bounty at the [ETH Denver Hackathon](https://www.ethdenver.com)

ailisp:
Modify to integrate the excellent work of fuzzy search made by @Markeljan into Gigs Board. Also allows to search result that both match search terms and with certain label. Close #66 

Preview:

https://near.social/#/bo.near/widget/gigs-board.pages.Feed?nearDevGovGigsContractAccountId=devgovgigs.near

### DIFF
--- a/src/gigs-board/components/community/CommunityHeader.jsx
+++ b/src/gigs-board/components/community/CommunityHeader.jsx
@@ -84,7 +84,6 @@ const BreadcrumbBold = styled.b`
 // but it's not there in near social, need write such style here
 return (
   <>
-
     <Header>
       <div aria-label="breadcrumb">
         <ol class="breadcrumb">
--- a/src/gigs-board/components/community/Layout.jsx
+++ b/src/gigs-board/components/community/Layout.jsx
@@ -86,9 +86,10 @@ const communities = {
 /* END_INCLUDE: "communities.jsx" */
 
 const Content = styled.div`
-{
-  padding: 0 32px;
-}`
+   {
+    padding: 0 32px;
+  }
+`;
 
 if (!props.label) {
   return (
--- a/src/gigs-board/components/layout/Banner.jsx
+++ b/src/gigs-board/components/layout/Banner.jsx
@@ -45,19 +45,19 @@ function href(widgetName, linkProps) {
 /* END_INCLUDE: "common.jsx" */
 
 const Banner = styled.div`
-{
-  height: 100px;
-  background: #232323;
-  margin-top: -24px;
-  overflow: hidden !important;
-  padding: 32px;
+   {
+    height: 100px;
+    background: #232323;
+    margin-top: -24px;
+    overflow: hidden !important;
+    padding: 32px;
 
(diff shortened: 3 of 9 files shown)


frol:
@ailisp Here are a couple of observations that I got playing with the preview link:

1. typing some long text (e.g. "mynearwallet") into the search box I noticed that the whole site get frozen and my typed text appears with a several seconds delay. Is there a chance to improve the situation?
2. Searching for "MNW" it says that nothing is found and instead falls back to search for "end" (well, totally not what I mean here)
3. Searching to "ZK" it also says that nothing is found and instead the search will be for "zkp"
4. The search only works for top-level posts, so replies are not indexed/displayed (I think it is fine)
5. The search downloads all the posts to index them (this is why I believe (4) is fine), so with the growing number of posts, we will hit the performance issues.

Overall, even with the current state, it is already a huge improvement, so I believe we can merge it and improve later.

Markeljan:
Awesome seeing this get integrated.  P.S. this widget was built by @p516entropy tagging and I will ping him as well.

ailisp:
@frol I can reproduce your observations. Will try fix 1 and 2. 2 looks like a bug of fuzzy search. There are posts with MNW wording but the search cannot find it and began to fuzzy search it. For 3 the search is set to search for not less than 3 letters, I can test it with 2 letters and see if the performance is okay. For 5, I have similar thought. Besides this one, looking at existing issues, a number of them would be benefit from having an indexer instead of doing in frontend and write indexing logics in smart contract.

ailisp:
Fixed 2 & 3. 2 is actually not a bug - all existing posts were not containing `mnw`, except comments, but comments are not shown in search result. So it attempts to treat mnw as a typo and did a fuzzy search. I added a post with `mnw` and now search `mnw` gives the result.

For 3, make it search words with 2+ characters. Now search for zk works.

For 1, I didn't figure out a better solution than existing. I was planned to have the search box only search when user press enter. But near-social viewer doesn't support having both `onKeydown`  (to check "Enter" is pressed) and `onChange` on an input - console log indicates only onKeydown was triggered. The other way is using a submit form, but near-social doesn't allow `form` element.

PR #94, opened 2023-04-14
integrate hottest post on the navbar

ailisp:
Integrate @p516entropy and @Markeljan's work of hottest posts on navbar. Also make it works together with search, filter by labels/authors.

Preview:
image

https://near.social/#/bo.near/widget/gigs-board.pages.Feed?recency=hot&nearDevGovGigsContractAccountId=devgovgigs.near

### DIFF
--- a/src/gigs-board/components/posts/List.jsx
+++ b/src/gigs-board/components/posts/List.jsx
@@ -49,7 +49,9 @@ function href(widgetName, linkProps) {
 }
 /* END_INCLUDE: "common.jsx" */
 
-console.log(props);
+initState({
+  period: "week",
+});
 
 function defaultRenderItem(postId, additionalProps) {
   if (!additionalProps) {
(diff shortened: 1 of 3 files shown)


src/gigs-board/components/posts/List.jsx
```diff
@@ -148,6 +150,76 @@ if (props.searchResult) {
...
+      postScore: getHotnessScore(post),
+    };
+  });
+  modifiedPosts.sort((a, b) => b.postScore - a.postScore);
+  return modifiedPosts.map((post) => post.id);
+};
+
+if (props.recency == "hot") {
+  postIds = findHottestsPosts(postIds, state.period);
+}
```
frol, line 221: Move here the conditions that we have above, that pull `postIds`, so we don't [fetch top-level posts](https://github.com/near/devgigsboard-widgets/pull/94/files#diff-39121cdc76ae3bae9ebc061f56ba14eb047d13a60474dc83e83edf13ce55d314R150) unnecessarily.

src/gigs-board/components/posts/List.jsx
```diff
@@ -260,13 +333,73 @@ const items = state.items ? state.items.slice(0, state.displayCount) : [];
...
+  <>
+    {Head}
+    <InfiniteScroll
+      pageStart={0}
+      loadMore={makeMoreItems}
+      hasMore={state.displayCount < state.items.length}
+      loader={loader}
+    >
+      {renderedItems}
+    </InfiniteScroll>
```
frol, line 403: Sometimes, there are no items on the list (e.g. "Hot 24h" is currently empty), let's display some meaningful text in that case.

src/gigs-board/components/posts/List.jsx
```diff
@@ -49,7 +49,9 @@ function href(widgetName, linkProps) {
 }
 /* END_INCLUDE: "common.jsx" */
 
-console.log(props);
+initState({
+  period: "week",
+});
```
frol, line 54: I would just do 3 different `recency`: `hot-24h`, `hot-week`, `hot-month` instead of using inner state, and this way there will be a way to link to the specific hot page, currently, there is no way to link to "Hot 24h" page as "Hot" page uses "week" by default.

frol (changes requested):

src/gigs-board/components/posts/List.jsx
ailisp: This can't be simply moved above. `findHottestsPosts` process on results of filtering/searching so it's a condition checked on all cases.

src/gigs-board/components/posts/List.jsx
ailisp: Good point, I also implemented the version as you suggested. However, the hot is like a way of sorting and it works together with search. But search is rely on state so the link of `recency=hot-24h` would be confusing, if user inputs with search term and shares the link it will become the hot without search. Therefore let's not do so before the search term is also in query string.

src/gigs-board/components/posts/List.jsx
```diff
@@ -162,7 +234,8 @@ const loader = (
 if (postIds === null) {
   return loader;
 }
-const initialItems = props.searchResult ? postIds : postIds.reverse();
+const initialItems =
+  props.searchResult || props.recency == "hot" ? postIds : postIds.reverse();
```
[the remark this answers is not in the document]
ailisp: Done!

src/gigs-board/components/posts/List.jsx
ailisp: Addressed!

src/gigs-board/components/posts/List.jsx
ailisp: I tried a few of ways to rewrite this if logic, but not able to figure out a better alternative. They are both more repetitive in combining the conditions and introduce more level of nestings that's less readable. Do you have a better way to combine filter, search and sorting (at frontend)? Eventually I'm lean for this approach: https://www.loom.com/share/77ab9a4acbf34e1aa65961dc70459a22l, that will also resolves the performance problem.

src/gigs-board/components/posts/List.jsx
frol: I see. Well, it goes wild already with all those combinations, and the lack of navigation API makes things even more confusing.

PR #101, opened 2023-04-18
Hide labels that user does not have permission to add when create and edit

ailisp:
Resolves https://github.com/near/devgigsboard-widgets/issues/10.

After this PR, labels that user does not have permission to add are hidden, when create and edit a post. Preview: https://near.social/#/bo.near/widget/gigs-board.pages.Feed?nearDevGovGigsContractAccountId=devgovgigs.near

And there is a warning if user without permission is trying to add such label:
image

Also, when user try to remove a label that he doesn't have a permission, it will now fail on UI and it will not be removed:

image

Previous to this PR it would not fail until submit the transaction:
Screenshot 2023-04-20 at 3 08 14 PM

Note that due to current near social viewer's limitation, it's not possible to add a lock icon to the protected labels like below, because related react-bootstrap-typeahead APIs are not exported:

Screenshot 2023-04-20 at 3 29 01 PM

### DIFF
--- a/src/gigs-board/components/posts/PostEditor.jsx
+++ b/src/gigs-board/components/posts/PostEditor.jsx
@@ -69,6 +69,7 @@ initState({
   token: props.token ?? "Near",
   supervisor: props.supervisor ?? "",
   githubLink: props.githubLink ?? "",
+  warning: "",
 });
 
 let fields = {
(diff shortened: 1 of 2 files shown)


src/gigs-board/components/posts/PostEditor.jsx
```diff
@@ -196,8 +196,15 @@ const setLabels = (labels) => {
...
 };
-const existingLabelStrings =
+let existingLabelStrings =
   Near.view(nearDevGovGigsContractAccountId, "get_all_labels") ?? [];
+existingLabelStrings = existingLabelStrings.filter((label) =>
+  Near.view(nearDevGovGigsContractAccountId, "is_allowed_to_use_labels", {
+    editor: context.accountId,
+    labels: [label],
+  })
+);
```
frol, line 206: Ouch, making 150+ view calls is extremely inefficient, especially as it does it on page load, not even at the point where users open the Post Editor. I would consider implementing a new method on the contract side similar to `is_allowed_to_use_labels`, but instead of returning a boolean, I would return a list of forbidden labels.

frol (changes requested):

ori-near:
> @ori-near Please edit one of my post, for example this one: https://near.social/#/bo.near/widget/gigs-board.pages.Post?id=112&nearDevGovGigsContractAccountId=devgovgigs.near
> 
> adding a `mnw` label (I don't have permission to add that label), so I can test implementation that I can't remove the label.

@ailisp – Done. Please note the `mnw` label didn't exist in the dropdown when I applied it.

ailisp:
@frol please take another look!

src/gigs-board/components/posts/PostEditor.jsx
```diff
@@ -186,18 +187,68 @@ const normalizeLabel = (label) =>
...
+    let removed = oldLabels.values().next().value;
+    Near.asyncView(
+      nearDevGovGigsContractAccountId,
+      "is_allowed_to_use_labels",
+      { editor: context.accountId, labels: [removed] }
+    ).then((allowed) => {
+      if (allowed) {
+        let labelStrings = labels.map((o) => {
+          return o.name;
+        });
```
frol, line 228:
```suggestion
        let labelStrings = labels.map(({ name }) => name);
```

Apply the same above and below.

P.S. What is the reason to call the variable `o`?

src/gigs-board/components/posts/PostEditor.jsx
```diff
@@ -208,12 +259,24 @@ const labelEditor = (
...
+      allowNew={(results, props) => {
+        return (
+          !existingLabelSet.has(props.text) &&
+          props.selected.filter((selected) => selected.name === props.text)
+            .length == 0 &&
+          Near.view(
+            nearDevGovGigsContractAccountId,
+            "is_allowed_to_use_labels",
+            { editor: context.accountId, labels: [props.text] }
+          )
```
frol, line 277: I feel like it may break since `Near.view` is not blocking, and on the first (non-cached) run it will return null/undefined. Do I miss something?

frol (approved):
The implementation is a bit hard to follow but looks fine to me now.

src/gigs-board/components/posts/PostEditor.jsx
ailisp: Experiments show this allowNew is invoked twice. Try type a new label, it shows from "No matches found" (non-cached) to "New selection xxxx" (Near.view resolves). If type label xxxx again, cache is hit, it immediately shows "New selection xxx".

Issue #81, opened 2023-04-06
[Communities] Enable permissions for communities – Smart Contract Side


ori-near:
**Problem**
Currently, after Community moved to smart contract from hard coded, only admins in DevDAO communities can create communities. Even moderators cannot edit them. This ticket is to introduce the appropriate permissions system for the communities feature on the smart contract side.

**User Stories**
- As a DevHub moderator, I need to have the permission to edit any community so that I can support our communities with any issues that arise.
- As a user with community moderator permissions, I need to have the permissions to edit the relevant items for my community so that I can make the necessary changes without relying on DevDAO developers.

**Acceptance Criteria**
- Moderators can edit any community
- Community admin can edit community, include add or remove another admin of this community

**Technical Implementation**
Since the label based permission system currently does require contract owner (devgovgigs.near) to change any permission rule, we'll come up with some simple straightforward solution:
- (Already in the contract) who already in Community Admin field can edit anything about the community, include add or remove admins
- (To implement in this ticket) moderators can edit any community, regardless of the admin settings

ori-near:
@nearmax @frol @ailisp – Please see this ticket for the proposed "admin" tab and let me know if you have any feedback or concerns. For this MPV, I am leaving the Jira/Kanban management out of scope, which we will solve in a separate ticket.

near-akaia-root:
@ori-near let's change the title to the following:

# [Communities] Store existing communities on chain & add permission control to enable data ownership for community members

near-akaia-root:
Let's also move it to "In progress"

Issue #78, opened 2023-04-06
[Communities] Sync Telegram Activity


ori-near:
**Problem**
Many of the Community Groups use Telegram and other social channels for discussions. However, our platform does not show real-time updates from those channels. This makes it hard for users to discover and keep track of conversations across the communities they care about. It also makes it hard for various roles/groups (e.g. Work Group members, DevDAO, etc) to coordinate monitoring effort of all channels, collect and aggregate feedback, and funnel it into the appropriate pipeline.

**Solution**
Implement a Telegram sync feature that enables each community page to show the latest activity from its corresponding Telegram channel. This will improve discoverability and engagement for the community, ultimately resulting in more contributions to the ecosystem.

**Acceptance Criteria**

- Integration with Telegram API to retrieve the latest activity from each channel
- Add admin capability for a user with admin permissions for a community to enable a "Telegram" tab and specify the various Telegram URLs
- If a community has Telegram enabled, display a new optional "Telegram" tab" along with the latest activity in real time
- If there are multiple telegram channels, separate them with a title and divider
- For MVP, please also populate these existing groups with the following Telegram channels:
  - Zero Knowledge: https://t.me/NearZeroKnowledge
  - Protocol: https://t.me/NEAR_Protocol_Community_Group
  - Tooling: https://t.me/NEAR_Tools_Community_Group
  - Contract Standards: https://t.me/nearnft

ori-near:
@nearmax / @vlad-kopcil – I discussed this ticket from @ailisp, and he had a question about how we want to store the data. We have a few potential options:

1. Fetching the latest feed (e.g. last 1000 messages) from Telegram and not storing them in the smart contract.
2. Storing the Telegram messages in a smart contract by creating a post / other data type for each Telegram message.
3. Storing the Telegram messages in a database/backend

Bo recommends we proceed with option #1. Please let us know if you have any concerns or alternatives.

nearmax:
In general, we should try avoiding any middleware or backend as much as possible, because DevDAO is not going to have large operations to be responsible for many services. I was thinking about two options:
* Store synced Telegram messages in SocialDB;
  * Pros:
    * It does not make the code of the main contact more complex.
    * The confirmation UX it simplified, because today when you call a method of a smart contract which is not SocialDB it always shows you a transaction confirmation window, while with SocialDB interactions it frequently skips it -- this will however become relevant only once we add a bi-directional sync.
[...]

PR #126, opened 2023-05-12
Community latest telegram messages

ailisp:
Address #78. This uses the official telegram widget (iframe) to display individual message and I setup an aws lambda service to get the latest 10 message IDs from telegram groups: https://github.com/ailisp/cg-msg-counter/blob/main/index.js. There is no frontend-only solution to retrieve the message IDs.

Preview: https://near.social/#/bo.near/widget/gigs-board.pages.Feed?nearDevGovGigsContractAccountId=devgovgigs.near

Note that, due to #81 is not there, this PR uses hard coded telegram groups.

@ori-near ~I have trouble in layout these telegram messages in a beautiful way (they're iframes from https://t.me/`), may you take a look at the css here?~

There is no pure css way to make every iframe as height as the content. The JS way isn't work because near social VM prohibit direct access to DOM element. Therefore, I will modify the service to return messages as json and render with CSS to try looks close to telegram's UI

### DIFF
--- a/src/gigs-board/pages/community/Telegram.jsx
+++ b/src/gigs-board/pages/community/Telegram.jsx
@@ -109,7 +109,6 @@ if (groupInfo === null || !groupInfo.ok) {
 }
 
 const messageIds = groupInfo.body.messageIds;
-console.log(messageIds);
 
 const Telegram = (
   <>
@@ -127,21 +126,6 @@ const Telegram = (
         minWidth: "100%",
       }}
     ></iframe>
-    {messageIds.map((id) => {
-      return (
-        <iframe
-          id={id.toString()}
-          iframeResizer={{ log: true, heightCalculationMethod: "scroll" }}
-          src={
-            "https://cg-msg-viewer.pages.dev/message?group=" +
-            group +
-            "&id=" +
-            id
-          }
-          style={{ width: "1px", minWidth: "100%" }}
-        />
-      );
-    })}
   </>
 );
 
--- a/src/gigs-board/pages/community/Telegram.jsx
+++ b/src/gigs-board/pages/community/Telegram.jsx
@@ -101,24 +101,16 @@ const community = communities[props.label];
 
 const group = community.telegram;
 
-const groupInfo = fetch(
-  `https://j96g3uepe0.execute-api.us-east-1.amazonaws.com/groups/${group}`
-);
-if (groupInfo === null || !groupInfo.ok) {
-  return "Loading ...";
-}
-
-const messageIds = groupInfo.body.messageIds;
-
 const Telegram = (
   <div>
     <iframe
-      iframeResizer={{ log: true }}
+      iframeResizer
       src={
         "https://j96g3uepe0.execute-api.us-east-1.amazonaws.com/groups-ui/" +
         group
       }
       frameborder="0"
+      // Required by iframeResizer
       style={{
         width: "1px",
         minWidth: "100%",
--- a/src/gigs-board/pages/community/Telegram.jsx
+++ b/src/gigs-board/pages/community/Telegram.jsx
@@ -97,40 +97,48 @@ if (!props.label) {
   );
 }
 
-const label = props.label;
+const community = communities[props.label];
 
-const discussionRequiredPosts =
-  Near.view(nearDevGovGigsContractAccountId, "get_posts_by_label", {
-    label,
-  }) ?? [];
+const group = community.telegram;
 
-const Discussions = (
-  <div>
-    <div class="row mb-2">
-      <div class="col text-center">
-        <small class="text-muted">
-          Required label:
-          <a href={href("Feed", { label })} key={label}>
-            <span class="badge text-bg-primary me-1">{label}</span>
-          </a>
-        </small>
+const groupInfo = fetch(
+  `https://j96g3uepe0.execute-api.us-east-1.amazonaws.com/groups/${group}`
+);
+if (groupInfo === null || !groupInfo.ok) {
+  return "Loading ...";
+}
+
+const messageIds = groupInfo.body.messageIds;
+
+const Telegram = (
+  <div class="">
+    {messageIds.map((i) => (
+      <div class="row">
+        <div class="col">
+          <iframe
+            id={"telegram-post-" + i}
+            key={"telegram-post-" + i}
+            src={"https://t.me/" + group + "/" + i + "?embed=1&userpic=true"}
+            scrolling="no"
+            style={{
+              overflow: "hidden",
+              colorScheme: "light dark",
+              border: "medium none",
+              minWidth: "100%",
+              minHeight: "100px",
+            }}
+            width="100%"
+            height="100%"
+            frameborder="0"
+          ></iframe>
+        </div>
       </div>
-    </div>
-    {widget("components.layout.Controls", {
-      labels: discussionsRequiredLabels,
-    })}
-    <div class="row">
-      <div class="col">
-        {discussionRequiredPosts.map((postId) =>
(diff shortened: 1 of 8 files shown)


ailisp:
So this problem is surprisingly challenging due to the limitation that combined by telegram API & near social. So far I have experimented the following approaches, but didn't get any of them work on the near social yet:

1. direct iframe of https://t.me message widget. The iframes are not layout well since css can't size the chat message according to its content. JS doesn't work because near social prevent DOM access.
2. host my static site to layout https://t.me message widget, and iframe mysite. However JS cannot cross origin access iframe's scrollHeight due to https://t.me's server setup
3. (current preview). host my static site to layout telegram widgets and iframe this static site in DevHub. The static site looks good, for example: https://cg-msg-viewer.pages.dev/?group=NearZeroKnowledge. But when load it in DevHub, it doesn't layout right again, due to near social limitation. So I have to get messages in server side and find a frontend library to render it. Frontend telegram alone doesn't work because it requires user login to even list public group messages.
4. So I use telegram nodejs library based on mtproto, it can efficiently give a list of messages in [TL Schema Format](https://core.telegram.org/constructor/message), for example: https://j96g3uepe0.execute-api.us-east-1.amazonaws.com/messages/cryptonear. Then I'm trying to find a frontend library to render it. 
5. Official telegram web version Z and version K are both use non standard (self made frontend framework) that are very challenging to integrate with the frontend
6. Telegram-react, I'm able to run the Chat Message component, but the message format it takes from is in [tdlib format](https://core.telegram.org/tdlib/docs/classtd_1_1td__api_1_1message.html), which is not compatible to TL Schema Format above. And I found the conversion is non trivial.
7. Then I try to modify the service in 4. Although tdlib is more highlevel than mtproto, it's more difficult to use. tdlib's user session tokens is challenging to work with. Also tdlib's doesn't have official nodeJS binding, only browser JS, and non official node JS binding libraries have trouble in building and/or working on aws lambda. 

I plan to move to tdlib in another language to implement 7, then I will get the messages in right format for 6, and make a component and host on cloudflare pages, then iframe it here in DevHub.

ailisp:
@frol Thanks! I'll create an issue on vm viewer.

> I believe that a page with content and two scrolls on it is better than no such page, so let's not go too deep into the custom rendering of Telegram messages.

Not sure if I read this correctly. Is this the resulting thing after integration integration iframe support?

> The current widget has reverse order of messages and uneven spacing

For ordering, it is intentional - want to display latest messages first to be consistent with posts. Do we want the last message last? @ori-near 

To clarify for others who come to discussion, the uneven spacing problem is the blocker we're trying to solve with all above comments.

ailisp:
I found near social vm has iframe-resizer: https://github.com/NearSocial/VM/pull/29. However the support is incomplete and buggy, for example this issue: https://github.com/NearSocial/VM/issues/32. I've experimented several configurations of iframe-resizer on parent and on containing iframe, non of them could correctly obtain the message iframe height and pass to the parent :thinking:

frol:
> > I believe that a page with content and two scrolls on it is better than no such page, so let's not go too deep into the custom rendering of Telegram messages.
>
> Not sure if I read this correctly. Is this the resulting thing after integration iframe support?

Yes, what I suggest is that we find a minimal working solution and run with it for now.

I also took a quick search for the ideal solution. I wish there is a standard widget for that, but Telegram Core only [offers widget for discussions](https://core.telegram.org/widgets/discussion) (a single message on a channel + comments) - this widget looks great, but it cannot preview a public group. [Web-clients](https://core.telegram.org/widgets/discussion) require authorization, and you cannot even preview a public group like https://web.telegram.org/k/#@neardev

ailisp:
> [Web-clients](https://core.telegram.org/widgets/discussion) require authorization, and you cannot even preview a public group like https://web.telegram.org/k/#@neardev

What I am suggested in 6 & 7 above is essentially storing authorization on a aws lambda service and expose public group messages, then a modified web clients to render the messages. This is a working aws lambda in action: https://j96g3uepe0.execute-api.us-east-1.amazonaws.com/messages/cryptonear,
but the message format (mtproto) doesn't fit telegram-react (tdlib). It fits the official telegram web version K, but version K is using a self-made frontend framework and quite challenging to work with. I feel the other way, change the service to output into tdlib way and modify telegram-react seems to take less effort.

frol:
@ailisp I wonder if [allowing more sandbox features](https://www.w3schools.com/tags/att_iframe_sandbox.asp) can fix the issue.

ailisp:
@frol great suggestion! Experiment with a local near social viewer/VM, allow-same-origin is all I need to fix this issue. Will create a PR.

ailisp:
I figure out an interesting way to render telegram messages well. Using an aws lambda function to crawl https://t.me/group_name/message-id, then take it as a div, combine divs to a single page dynamically, and iframe this dynamic page in near social. 

near social can correctly layout divs in an iframed page, but not iframe in iframe, as a result this way looks good:

image

frol:
@ailisp I checked your preview link from the PR description and the background is black (is it intended?) and the font color in the messages is white (impossible to read messages since the background there is white):

image

https://near.social/#/bo.near/widget/gigs-board.pages.community.Telegram?label=zero-knowledge&tab=Telegram&nearDevGovGigsContractAccountId=devgovgigs.near

src/gigs-board/pages/community/Telegram.jsx
```diff
@@ -0,0 +1,134 @@
...
+if (groupInfo === null || !groupInfo.ok) {
+  return "Loading ...";
+}
+
+const messageIds = groupInfo.body.messageIds;
+
+const Telegram = (
+  <div>
+    <iframe
+      iframeResizer={{ log: true }}
```
frol, line 116: Should we disable `log` now?

src/gigs-board/pages/community/Telegram.jsx
```diff
@@ -0,0 +1,134 @@
...
+      src={
+        "https://j96g3uepe0.execute-api.us-east-1.amazonaws.com/groups-ui/" +
+        group
+      }
+      frameborder="0"
+      style={{
+        width: "1px",
+        minWidth: "100%",
```
frol, line 124: Can you leave a comment about this quirk? Do we still need it, by the way?

src/gigs-board/pages/community/Telegram.jsx
```diff
@@ -0,0 +1,134 @@
...
+const group = community.telegram;
+
+const groupInfo = fetch(
+  `https://j96g3uepe0.execute-api.us-east-1.amazonaws.com/groups/${group}`
+);
+if (groupInfo === null || !groupInfo.ok) {
+  return "Loading ...";
+}
+
+const messageIds = groupInfo.body.messageIds;
```
frol, line 111: It seems that `messageIds` and `groupInfo` is not used. Let's delete dead code

frol (approved):
There are some styling issues and minor comments to address, so I optimistically approve this PR, so you can merge this PR once all of the comments are addressed.

src/gigs-board/pages/community/Telegram.jsx
ailisp: Yes, in this non-nested case, iframe resizer works, with these required by iframe-resizer

ailisp:
> @ailisp I checked your preview link from the PR description and the background is black (is it intended?) and the font color in the messages is white (impossible to read messages since the background there is white):
> image
> 
> https://near.social/#/bo.near/widget/gigs-board.pages.community.Telegram?label=zero-knowledge&tab=Telegram&nearDevGovGigsContractAccountId=devgovgigs.near

Looks like you are in dark mode and the css is changing to black background in dark mode. I'll fixed it

PR #148, opened 2023-06-07
add contribution process for who working on a regular basis

ailisp:
Publish the contribution process and link from readme


README.md
```diff
@@ -20,6 +20,8 @@ We communicate primarily over GitHub. We recommend configuring your [notificatio
 
 [Learn more](https://github.com/near/devgigsboard-widgets/blob/main/docs/how-we-work.md) about how we work asynchronously.  
 
+**Please follow [this process](https://github.com/near/devgigsboard-widgets/blob/main/docs/contribution.md) if you work on a regular basis.**
```
frol, line 23: Most of the things are relevant for the general [CONTRIBUTING](https://github.com/near/neardevhub-widgets/blob/main/CONTRIBUTING.md) guide. It would be ideal to merge the two as I believe there should not be an artificial barrier to one-off contributors, and the described process works will work fine except the escalation through the private Telegram group, which I hope won't happen often and it is probably better to escalate through @ori-near on GitHub.

frol (approved):
It is straight to the point which is great, but I am concerned about duplication a bit as it hurts discoverability and may potentially make contradicting statements or outdated instructions.

README.md
ori-near: @frol – I think I addressed your concerns with the new PRs that I added. Can you please review and let me know if you have any other questions? I think we can cancel this PR and look at the newest ones.

ailisp:
Closed as #153 #152

PR #153, opened 2023-06-10
Update how-we-work.md

ori-near:
Added more clarity on how we can work most effectively together based on recent learnings. New sections include: Agreements, developer workflow, and communication.


ailisp (approved):
Looks great!

Issue #170, opened 2023-06-19
Center featured community


ailisp:
**Problem** 
Featured community right side has extra margins than the left side on homepage.

**Acceptance Criteria**
Make them centered on page

PR #190, opened 2023-06-26
fix: margin right and left to be equal

Tguntenaar:
fixed [#170](https://github.com/near/neardevhub-widgets/issues/170)


Tguntenaar:
Screenshot 2023-06-26 at 22 50 58

frol (approved):

PR #197, opened 2023-07-03
Save post progress and prevent data loss on cancel

### LINKED ISSUE #87: [Create Post] Save post progress and prevent data loss on cancel
**Problem**
Currently, when a user creates a new post on our platform and clicks submit, they may lose all their work if something goes wrong on the subsequent screen. We've received several complaints from users about this, as this is causing them frustration and wasted time. 

Example user flow:
- Create a new post (e.g. Idea) and enter all the information into the fields
- Click submit, which prompts the user with the Confirm Transaction pop-up message
- Click confirm, which redirects the user to the Approve Transaction screen
- Click cancel, which redirects the user to the homepage and causes data loss

https://user-images.githubusercontent.com/110252255/230693071-86a7a0f5-e4e9-46ba-9ce7-129ffd046a34.mp4

**User Story**
As a user, I want to have the ability to save my post progress and return to my draft if I accidentally navigate away or close the window, so that I don't lose all my work and can easily pick up where I left off.

**Acceptance Criteria**
- When a creates a new post, clicks submit, and then confirm, if the user clicks cancel or goes back in the browser from the Approve Transaction screen, the user should still see their draft
- Users can submit the saved draft when they are ready to publish it

petersalomonsen:
Work in progress

Preview URL: https://near.org/devgovgigs.petersalomonsen.near/widget/gigs-board.pages.Feed

- On submit, the editor contents will be saved in `Storage` ( private for the widget )
- Not confirming or cancelling the transaction in the wallet will restore the draft when returning to the page
- If posting a reply or editing an existing post, the reply/edited post is expanded and put to edit mode
- The presence of `transactionHashes` in the query string will indicate that the post was submitted successfully and the draft can be cleared from the local storage.

In order to make sure that the draft is cleared upon successful submission, a confirmation is introduced with a link back to the feed. Going back to the feed will ensure that `transactionHashes` is cleared from the query string. Ideally going back to the feed should happen immediately without an intermediate manual step, but there is no way to day in NEAR BOS to automatically redirect.

image

resolves #87

### DIFF
--- a/build.js
+++ b/build.js
@@ -46,6 +46,12 @@ replaceInFiles(options)
       .readFileSync("./core/lib/autocomplete.js", "utf8")
       .toString()}/* END_INCLUDE: "core/lib/autocomplete" */`,
   })
+  .pipe({
+    from: /\/\* INCLUDE: "core\/lib\/draftstate" \*\/.*\/\* END_INCLUDE: "core\/lib\/draftstate" \*\//gms,
+    to: `/* INCLUDE: "core/lib/draftstate" */\n${fs
+      .readFileSync("./core/lib/draftstate.js", "utf8")
+      .toString()}/* END_INCLUDE: "core/lib/draftstate" */`,
+  })
   .then(({ changedFiles, countOfMatchesByPaths }) => {
     console.log("DONE");
   })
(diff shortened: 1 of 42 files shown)


ailisp (approved):
Looks great! A minor problem:
This "back to feed", i think it could be larger and more spacing from the search bar, to make it more obvious:

image

I'll merge this PR to unblock your another one.

ori-near:
Hi @petersalomonsen – thanks for working on this. This feature unforutnately introduced a critical bug: 
https://github.com/near/neardevhub-widgets/issues/199

Since you're out, I asked @near-akaia-root to work on it. But feel free to share any insights if you have any.

CC: @ailisp @frol

petersalomonsen:
Check out my reply here: https://github.com/near/neardevhub-widgets/issues/199#issuecomment-1631768052

PR #201, opened 2023-07-12
fix: Set minimum height for app layout

near-akaia-root:
Some elements may be cropped due to app layout height not being fixed.

Example:
imagen

With the given fix:
imagen

Page shown on the screenshots:
https://near.org/devgovgigs.near/widget/gigs-board.pages.Post?id=696

Fix preview:
https://near.org/root.akaia.near/widget/gigs-board.pages.Post?id=696&nearDevGovGigsContractAccountId=devgovgigs.near

### DIFF
--- a/src/gigs-board/components/layout/Page.jsx
+++ b/src/gigs-board/components/layout/Page.jsx
@@ -53,9 +53,9 @@ function href(widgetName, linkProps) {
 /* END_INCLUDE: "common.jsx" */
 
 return (
-  <>
+  <div className="h-100" style={{ minHeight: "86vh" }}>
     {widget("components.layout.app-header", {})}
     {props.header}
     {props.children}
-  </>
+  </div>
 );


src/gigs-board/components/layout/Page.jsx
```diff
@@ -53,9 +53,9 @@ function href(widgetName, linkProps) {
 /* END_INCLUDE: "common.jsx" */
 
 return (
-  <>
+  <div className="h-100" style={{ minHeight: "86vh" }}>
```
frol, line 56: Where is 86vh coming from? I see it partially helping, but it seems we need to have a better solution in the future.

frol (approved):
As I mentioned, I see that it partially addresses the problem

src/gigs-board/components/layout/Page.jsx
near-akaia-root: It's the minimum value that works

Issue #142, opened 2023-05-30
Create solution post & solution reply are different 


ori-near:
With the recent launch of the new post creation UX (https://github.com/near/neardevhub-widgets/issues/20), currently the fields for the create solution are different from the reply solution This leads to an inconsistent UI and user experience. For example, right now a user is required to select the currency from a dropdown in the Solution post, which includes the option for USDC. However, the Solution reply only exposes a field, and a user gets a warning when they enter USDC.

**To Reproduce**
1. Go to DevHub
2. Click on Submit Idea / Solution, then Solution
3. User sees the following fields:
  - Title (text field)
  - Description (large text field)
  - Labels (type dropdown)
  - Are you seeking funding? (Radio button yes or no)
  - When user selects yes, they see:
    - Currency (dropdown with three options: NEAR, USDC, USD)
    - Requested amount (number only text field)
    - Requested sponsor (text field)
4. Navigates to any existing post on the feed and click Reply, then Solution
5. User sees the following fields, which are different for the ones shown in step 3:
    - Labels (type dropdown)
    - Title (text field)
    - Description (large text field)

**Expected behavior**
- User should see the same fields & preview in the solution reply as shown in the create solution post
- User should be able to select USDC and proceed successfully without a warning message

**Screenshots**
<img width="1580" alt="Screenshot 2023-05-30 at 3 15 59 PM" src="https://github.com/near/neardevhub-widgets/assets/110252255/fb2327f5-f13c-48c2-ace1-d48dafe68061">

Tguntenaar:
I would like to tackle this

near-akaia-root:
Assigned @Tguntenaar

Tguntenaar:
Currently in progress

Tguntenaar:
I've finished the solution however it shows some weird behaviour rendering the input fields multiple times on state change. Still have to work on that before I make a PR. Also waiting @ori-near his reply what to do with the sponsorship option.

Tguntenaar:
> Hi @Tguntenaar – I believe we already discussed this on our call a couple of weeks ago. Please let me know if you're still blocked.

Not blocked! Created a PR for it yesterday as well.

ori-near:
Hi @Tguntenaar – I see that we closed this ticket – but I still see the bug where the solution reply is not the same as the solution create. Do we need to reopen this?

CC: @ailisp

ailisp:
> Do we need to reopen this?

Yes

[...]

PR #205, opened 2023-07-16
142 solution reply post

Tguntenaar:
Solves issue #142

[Preview version](https://near.org/thomasguntenaar.near/widget/gigs-board.pages.Feed)

-User sees the same fields & preview in the solution reply as shown in the create solution post
-User is able to select USDC and proceed successfully without a warning message


ailisp:
Hmm when I try reply from your preview, it's still not showing USDC:


image

Did you deploy something else that overrides this branch?

Tguntenaar:
@ailispI noticed that the previous preview was accidentally overwritten, I apologize for the late reaction. Here's the updated preview for your review [preview](https://near.org/thomasguntenaar.near/widget/gigs-board.pages.Feed)

src/gigs-board/entity/post/PostEditor.jsx
```diff
@@ -110,15 +113,9 @@ if (!state.draftStateApplied && props.draftState) {
...
+  Submission: ["name", "description", "fund_raising"],
   Attestation: ["name", "description"],
-  Sponsorship: [
-    "name",
-    "description",
-    "amount",
-    "sponsorship_token",
-    "supervisor",
-  ],
+  Sponsorship: ["name", "description", "fund_raising"],
```
ailisp, line 118: Is this intentional? I saw you proposed to remove this in the issue, but @ori-near hasn't reply so is this decided.

src/gigs-board/entity/post/PostEditor.jsx
```diff
@@ -142,7 +139,12 @@ const onSubmit = () => {
     },
     Submission: {
       name: state.name,
-      description: state.description,
+      description: generateDescription(
+        state.description,
+        state.amount,
+        state.token,
+        state.supervisor
+      ),
```
ailisp, line 147: If really to remove fields from sponsorship, below Sponsorship:
```
 Sponsorship: {
      name: state.name,
      description: state.description,
      amount: state.amount,
      sponsorship_token: state.token,
      supervisor: state.supervisor,
      sponsorship_version: "V1",
    },
```

 should also update, and call contract to create sponsorship should also update. I suggest to consider & address sponsorship in a future PR. in this PR, only do fix #142.

ailisp (commented):
Preview looks good! but sponsorship I don't recommend to change in this PR

src/gigs-board/entity/post/PostEditor.jsx
Tguntenaar: Yes I spoke with @ori-near on a call about this.

Issue #213, opened 2023-07-27
Clicking on Submit Idea / Solution opens up Solution post that misses the funding field


ori-near:
**Describe the bug**
When a user clicks on the "Submit Idea / Solution" create button, the second Solution post type loads by default without the funding fields. Additionally, the preview shows a ":" instead of the expected solution preview. There are two problems with this: One, the default view should show the idea type, since it's first. Second, the missing funding fields may lead a user to not including funding details.

**To Reproduce**
1. Go to neardevhub.org
2. Click on Submit Idea / Solution
3. See solution post without the funding fields and the incorrect preview: 
![Screenshot 2023-07-27 at 2 16 53 PM](https://github.com/near/neardevhub-widgets/assets/110252255/63f334e8-811d-424b-ad70-2f0ed70a0c2c)
4. The user can only see the correct fully loaded solution post by clicking on the Solution button again at the top.

**Expected behavior**
The idea post should load first. And the solution post should load correctly:
![Screenshot 2023-07-27 at 2 18 43 PM](https://github.com/near/neardevhub-widgets/assets/110252255/a4f4e31e-c007-4415-84cf-4f824d2992d5)

elliotBraem:
I can pick this one up @ori-near

ailisp:
@elliotBraem I'll quickly fix this by revert a PR

Issue #210, opened 2023-07-26
Sponsorship reply form confuses users with "Sponsor" and duplicate information in preview


frol:
**Describe the bug**

Replying with sponsorship produces unnecessary "Sponsored amount" and "Sponsor" in addition to the correct "Maximum amount" and "Supervisor" fields in **the preview**:

<img width="906" alt="image" src="https://github.com/near/neardevhub-widgets/assets/304265/bed0a305-cee7-4b5e-b5b7-0d389388d179">

Note: this is only an issue with the preview.

**To Reproduce**

1. Go to any post (https://near.social/#/devgovgigs.near/widget/gigs-board.pages.Post?id=842)
2. Click on "Reply" -> "Sponsorship"
3. Select "Yes" in "Are you funding their solution?"
4. Fill out the "Sponsored amount" and "Sponsor" (by the way, "Sponsor" is invalid in this context, it should be "Supervisor")

**Expected behavior**

1. The preview should only have "Maximum amount" and "Supervisor"  fields
2. "Sponsor" should be renamed to "Supervisor" in the Sponsorship reply form

ori-near:
Thanks for filing this @frol. I surfaced a similar concern to @Tguntenaar, as I believe this bug was introduced with https://github.com/near/neardevhub-widgets/issues/142. 

This bug introduced this same fields from the solution post in the sponsorship post:
![Screenshot 2023-07-27 at 2 29 28 PM](https://github.com/near/neardevhub-widgets/assets/110252255/92bcf16f-0cff-4b8d-9f2c-2cbb691f8f1d)

Please restore the original fields that the sponsorship pos had: Maximum Amount and Supervisor, remove the "Are you funding their solution" section, and make sure the preview renders correctly.

PR #215, opened 2023-07-28
Address #142

Tguntenaar:
_Resolves #142_
Address #142 and make sure not to introduce #210 and #213.

Expected behavior

1. User should see the same fields & preview in the solution reply as shown in the create solution post
2. User should be able to select ~USDC~ USDT and proceed successfully without a warning message

[Testnet Preview](https://test.near.social/thomasguntenaar.testnet/widget/gigs-board.pages.Feed)
And a  mainnet [preview](https://near.org/thomasguntenaar.near/widget/gigs-board.pages.Feed)


src/gigs-board/pages/Create.jsx
```diff
@@ -89,18 +89,18 @@ const labels = labelStrings.map((s) => {
...
-  //
+
   author_id: context.accountId,
   // Should be a list of objects with field "name".
   labels,
   // Should be a list of labels as strings.
   // Both of the label structures should be modified together.
   labelStrings,
-  postType: "Idea",
+  postType: "",
```
ailisp, line 99: This should be unchanged otherwise it cause bug #213

ailisp:
PR looks good overall. The description isn't accurate, this PR intends to address #142 without introduce bugs. The acceptance criteria is
> Expected behavior

 > User should see the same fields & preview in the solution reply as shown in the create solution post
    User should be able to select USDC and proceed successfully without a warning message

As i tested these criteria are met, so just ensure address above comments to avoid introduce a bug.

ailisp (changes requested):
Please address above comment to avoid introduce #213

ailisp:
@Tguntenaar I just update the title and acceptance criteria to make this PR's purpose accurate

ailisp:
Reply solution post doesn't include the "yes and no" section. Also please merge main branch and resolve conflict. Now main branch has only USDT and NEAR as currency.

Tguntenaar:
@ailisp The preview is probably showing the wrong version. Something with the bos cli blocks me from pushing the components to mainnet.

ailisp:
There are some bugs during my testing, list here for checking after fix:
- [ ] reply with sponsorship type, specify NEAR token gives this error:
image
- [ ] reply with solution type, specify USDT token renders preview and result post with "[object object]:
image

Below are correct cases, but need ensure it's still work after fix the above two issues.
- [ ] reply with sponsorship type, specify USDT token works fine. 
- [ ] reply with solution type, specify NEAR token works fine.
- [ ] create solution post with NEAR or USDT token work fine.

Tguntenaar:
- [x] reply with sponsorship type, specify NEAR token no longer gives error
- [x] reply with solution type, specify USDT token renders preview and result post correct

Below are correct cases, but need ensure it's still work after fix the above two issues.

- [ ] reply with sponsorship type, specify USDT token works fine.
- [ ] reply with solution type, specify NEAR token works fine.
- [ ] create solution post with NEAR or USDT token work fine.

Tguntenaar:
Hi @ailisp,

Finally tested all cases, they all work. I think it is ready to merge.

ailisp:
@Tguntenaar Mainnet preview no longer works, can you push a new one?

Testnet preview has some bugs:
- reply with a solution, when requested sponsor is empty, the requested fund doesn't show up:
image

- reply with a solution, enter the requested sponsor, and leave token as default (NEAR), but the transaction shows USDT:
image

Tguntenaar:
Hi @ailisp,

> reply with a solution, when requested sponsor is empty, the requested fund doesn't show up

This was intented it is the same behaviour as the Create widget. (Create a solution post and remove the requested sponsorship field) We could change it in both places. What would be the desired behaviour? A warning before submission that sponsorship is required or did you have something else in mind. 

> reply with a solution, enter the requested sponsor, and leave token as default (NEAR), but the transaction shows USDT

That is strange.. I didn't and don't experience this. It is probably caused by the snapshot of the state. I will try to reproduce the behaviour and come up with a solution.

ailisp:
OK! Sounds great then. I also tested on mainnet and the NEAR shown as USDT problem doesn't exist. For sponsorship empty it's not a new problem introduced in this PR, we might want to address in a separate ticket. Now all tests work fine. Could you please resolve conflict so we can merge it?

ailisp (approved):

PR #216, opened 2023-07-29
fix: confirmation only for add/edit post

### LINKED ISSUE #209: The post submitted successfully message is sometimes inaccurate
**Describe the bug**
Currently, when a user performs various actions (editing a post, creating a post, liking a post, etc) that forces the confirm transaction screen, they are redirect to a "Post submitted successfullly. Back to post" message screen. However, the user is not always performing a post submission action, so the confirm message is sometimes inaccurate, leading to confusion.

Post submitted successfullly. Back to post.

**To Reproduce**
1. Take any action (e.g. like a post, edit a post, etc)
2. Click on confirm transaction screen, then approve
3. See error

**Expected behavior**
Either change the error message to reflect the action the user took. Or make it generic (e.g. Action completed successfully). 

**Screenshots**
Screenshot 2023-07-24 at 7 21 32 AM

petersalomonsen:
This fix will check the method name of the returned transaction and only show the post confirmation screen if it is edit_post or add_post. This means that for a `like`, it will just return to the feed, and not show any confirmation screen. Also the confirmation screen now shows that post is added or edited.

Preview URL: https://near.org/devgovgigs.petersalomonsen.near/widget/gigs-board.pages.Feed?nearDevGovGigsContractAccountId=devgovgigs.petersalomonsen.near

fixes #209

### DIFF
--- a/module/core/lib/draftstate.js
+++ b/module/core/lib/draftstate.js
@@ -1,6 +1,27 @@
 const DRAFT_STATE_STORAGE_KEY = "POST_DRAFT_STATE";
+let is_edit_or_add_post_transaction = false;
 if (props.transactionHashes) {
-  Storage.privateSet(DRAFT_STATE_STORAGE_KEY, undefined);
+  const transaction = fetch("https://rpc.mainnet.near.org", {
+    method: "POST",
+    headers: {
+      "content-type": "application/json",
+    },
+    body: JSON.stringify({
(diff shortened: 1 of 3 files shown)


petersalomonsen:
@ori-near You might also want to review this, since you reported the bug.

ailisp (approved):
Looks good and test works to me!

Issue #172, opened 2023-06-19
[Communities] Implement form validation


ailisp:
**Problem**
Currently, a user is able to submit various inputs that are not allowed due to a lack of front end validation. This may result in error messages and poor user experience.

**User Story**
As a user interacting with the various forms on DevHub, I want to know in-real time if I am entering fields that are not allowed, so that I can have a smooth and error-free user experience once I click submit/next etc.

**Acceptance Criteria**

On the community create page:
- For all required fields (e.g. Name, Description, Handle, Tag, Admins) – a user should not be allowed to continue without filling them out. Also make the bio field under about required (add a red * next to Bio) and apply the same required behavior.
- For all fields with a maximum length requirement, form should prevent user from typing more characters. For example, for the Twitter Handle field, a user should not be allowed to enter more than the maximum 60 character limit limit.
- For URL handle and tag, do not allow commas or spaces.
- The website field show only allow to paste a valid URL
- If a user sets a name for wiki page, they must not leave content blank

ori-near:
Hi @ailisp Can you please re-esimate this ticket? And let me know if I should break this into two tickets (e.g. one for form validation, the other for improving the page access control)? Thank you.

ailisp:
@ori-near a week's estimate should be a safe upperbound. 
> Improve the UI experience for entering additional admins (see options below)

takes at most 3 days and remaining tasks should take at most 2 days

ori-near:
Hi @ailisp – thanks so much! I ended up splitting the https://github.com/near/neardevhub-widgets/issues/241 into a separate story since they seem parallelizable.

espensimensen:
@ori-near I can look into this ticket

PR #218, opened 2023-07-31
Simplified community creation flow and boards MVP

### LINKED ISSUE #156: [Communities] Enable Board MVP and enhance Create Community flow
**Problem**
Some users run into issues with creating communities after spending significant time on an extensive form. Additionally, communities lack an effective method to organize posts according to their workflows, making it difficult to visualize progress. To address these issues, we need to enhance the community creation experience and introduce a basic customizable Kanban board feature to organize posts.

**User Story**
As a user, I want a smooth community creation process without error and the ability to organize posts for my community to enhance visibility.

**Acceptance Criteria**

**1. Creating Communities**
- [ ] Clicking the "Create Community" button displays a concise form above the community list
- [ ] The form section is called "Community Information" and it display critical fields: Name, Description, URL Handle (Changed from Handle), and Tag (allow up to 30 characters).
- [ ] The form displays "Launch" and "Cancel" buttons. A user must enter all fields in order to click "Launch" and proceed.
- [ ] While the form is active, the "Create Community" button is hidden. Canceling the form restores the button.
- [ ] Once a user enters the required fields, clicks the "Launch" button and proceeds with the transaction confirmation, the form disappears and the new community appears at the end or beginning of the communities page
- [ ] By default, the user is set as the default admin and the placeholder images are set as the logo and banner

**2. Viewing Community (After Creation)**
- [ ] Clicking the newly created community reveals:
  - [ ] Top header contains "Configure Community" button for settings
  - [ ] Navigation tabs: Activity, Teams, GitHub, and Board (need to remove old sponsorships page)
  - [ ] Sidebar description by default contains placeholder text: "This is a sample text about your community. Edit this in the configure community page." Remove the label or anything else. Only text description should show.
Note: Group moderatos section will be updated as part of https://github.com/near/neardevhub-widgets/issues/241

**3. Configuring Board Tab**
- [ ] On the Board tab, all user see a similar message to the existing GitHub tab: "This community does not have a board set up."
- [ ] Admins will see a "Create Board" button below the message. Clicking on it leads to a configuration form, similar to the GitHub tab.
- [ ] Admins can configure the page name and enable up to 10 columns in the Kanban board
- [ ] Adnubs can choose which fields to show/hide in the Kanban cards (e.g. post title, post author, post type, post labels, # replies)
- [ ] Admin can customize column name, description, and filtering criteria
- [ ] Filtering criteria should display only DevHub posts with specific labels in each column

**4. Configuring Community Page**
- [ ] Admin has the ability to edit all critical fields that were displayed in the create flow.
- [ ] Admin has the ability to edit all other sections that were not displayed in the create flow (e.g. About, Access Control, Wiki Page 1, Wiki Page2)

carina-akaia:
Closes #156 

## Preview

- [Mainnet](https://near.social/carina.akaia.near/widget/gigs-board.pages.communities?nearDevGovGigsContractAccountId=devgovgigs.near)
- [Testnet, frontend + contract](https://test.near.social/1337hub.testnet/widget/gigs-board.pages.communities)


src/gigs-board/entity/project/kanban-view.jsx
```diff
@@ -0,0 +1,346 @@
...
+    ).reverse();
+
+    return {
+      ...registry,
+
+      [columnId]: {
+        ...column,
+
+        postIds:
+          tags.required.length > 0
```
ailisp, line 182: Where is the `tags` defined? I don't see it and there seems no check to format of `tags` before using it

src/gigs-board/entity/project/kanban-view.jsx
```diff
@@ -0,0 +1,346 @@
...
+
+  const columns = configToColumns(configuration);
+
+  return (
+    <div className="d-flex flex-column gap-4">
+      <div
+        className={"d-flex justify-content-end gap-3 p-3 rounded-4"}
+        style={{ backgroundColor: "#181818" }}
+      >
+        {(link ?? null) !== null ? (
```
ailisp, line 219: Looks weird. `link ?? null` checks if link is null or undefined. and set it to null if so. And then you check whether it is not null. If so display something, if not return null. Let's try to simplify this code.

pseudo code:
```
if link is null or undefined
  link = null
if link is not null
  return <>...</>
else
  return null
```

Simplify:
```
if link is null or undefined => set link to null => return null
else => return <>...</>
```

Simplify:
```
link ? <>...</> : null
```

src/gigs-board/entity/project/kanban-view.jsx
```diff
@@ -0,0 +1,346 @@
...
+          <>
+            {false && // Temporarily disabled
```
ailisp, line 221: recommend to put TODO, to quickly filter everything without missing

src/gigs-board/pages/Create.jsx
```diff
@@ -471,7 +471,7 @@ function generateDescription(text, amount, token, supervisor) {
 
 return (
   <div class="bg-light d-flex flex-column flex-grow-1">
-    {widget("components.layout.app-header")}
+    {widget("components.organism.app-header")}
```
ailisp, line 474: I just realized of one potential fail path with moving files: Suppose one file is moved but not all paths are updated, it can work ok because of historical deployment. But when deploy to new it can fail. So double check all renamed widget reference are updated, and it would be best to deploy everything to a new account before merge.

src/gigs-board/pages/project.jsx
```diff
@@ -0,0 +1,376 @@
...
+                  tabindex="0"
+                  key={metadata.id}
+                >
+                  {widget(
+                    permissions.can_configure
+                      ? "feature.project.view-configurator"
+                      : ["entity.project", metadata.kind].join("."),
+                    {
+                      link: [
+                        "https://near.social",
```
ailisp, line 342: let's not hard coded the gateway

carina-akaia: @ailisp but is there another way?

carina-akaia:
> Above screen looks empty, can we make the first view to be the default view, when user enters the project?

Yes, I'll do it, good catch

> Could you add az233.testnet?

Done

carina-akaia:
@ailisp 

> Above screen looks empty, can we make the first view to be the default view, when user enters the project?

Yes, I'll do it, good catch

> Could you add az233.testnet?

Done

carina-akaia:
> Overall this PR looks fine and thanks for your hard work! For future one, please avoid refactoring in a large feature PR, because the code for large feature is a big deal. Refactoring in current project size without tests is error prune, we should not mix it with already-challenging feature implementation.

In young projects born in a week, big updates often require big refactoring, in order to keep the overall architecture consistent, and it's hard to preliminary tell how the result would look like. But I'm trying to offload changes in the project's core as much as possible, and hopefully I'll establish a a straightforward process for this matter 🙏

src/gigs-board/entity/project/kanban-view.jsx
carina-akaia: I just applied even more straightforward `typeof link === "string" ?`. But in some cases, `(x ?? null) !== null` is a vital workaround for dealing with JS's implicit type conversion, which I wouldn't recommend relying on.

ori-near:
Hi @carina-akaia – thanks so much for all your hard work on this! The [Testnet, frontend + contract](https://test.near.social/1337hub.testnet/widget/gigs-board.pages.community.activity?handle=devhub-test) link didn't work for me (I just see a Loading... message), so I just tested on [Mainnet](https://near.social/carina.akaia.near/widget/gigs-board.pages.community.activity?handle=devhub-test&nearDevGovGigsContractAccountId=devgovgigs.near). Please see my feedback below, which includes a mix of bugs as well some suggested improvements to basic UI form fields. Please let me know if you have any questions and if there are items you think we should move to a separate ticket – but I think many are within scope. Thanks so much!

## Feedback

1. **Create Communities:** On Create Communities: Change tag 0/20 character limit to 0/30 character. 
2. **Create Communities:**  After entering the first field text into the first Name field, I clicked the tab button on my keyboard expecting my selection to advance to the next field (description). However, it seems like the form skipped from Name to Tag to Description to URL handle. Can we make sure this order is consistent?
3. **Create Communities:**  The Launch button flashes intermittently as I type characters into some of the fields. Ideally it doesn't flash.
4. **Create Communities:**  The Confirm Transaction Screen shows additional inputs that were not shown in the previous screen (bio_markdown, logo_url, banner_url). I think this is confusing to the discerning user because they may not know what this means. Not sure if we have an elegant way to solve this (e.g. maybe have a default value that only changes into smart contract once it's configured or something). Current solution is not ideal if there's a quick fix.
5. **Create Communities:**  The Confirm Transaction Screen placeholder text shows "\n" – can we remove it?
6. **Create Communities:**  When I entered more than the allowed character limit, I got this error message in the approve transaction screen: `{"index":0,"kind":{"ExecutionError":"Smart contract panicked: panicked at 'Community description must contain from 6 to 60 characters', src/community/mod.rs:100:13"}}` – This is not ideal, but I understand if we have to solve this in the separate form validation ticket.
7. **Create Communities:**  After launching the community, it appeared in the middle of the list of communities. This is confusing because it was hard to locate, and initially I thought the community wasn't created. I thought Carina mentioned that it would appear at the top. But if easier, we can just show it at the bottom of the page, similar to the previous placement of new communities.
8. **Viewing Community (After Creation):** Can you please change the styling of the Configure Community and Share button to reflect what was there before? I don't think we need to draw so much attention to these.
9. **Viewing Community (After Creation):**Sidebar shows up as empty with the community tag (clicking on it leads to an error page). The sidebar should only show the sample text.
10. **Configuring Community Page:** Same issue as 2 for the Community information section.
11. **Configuring Board Tab:** After entering all the fields, I am not actually able to get any cards to display. Instead I get an error message   
Screenshot 2023-08-16 at 5 40 51 PM

13. **Configuring Board Tab:** Can we update the Kanban board configuration fields as follows:
  - Change "Name" field to "Board title" and the placeholder "NEAR Protocol NEPs" to "Enter board title."
  - Change "All the tags MUST be presented in posts" to "Enter tags you want to include. Posts with these tags will display:" and the placeholder "near-protocl-neps," to "tag1, tag2"
  - Change "All the tags MUST NOT be presented in posts" to "Enter tags you want to exclude. Posts with these tags will not show." and the placeholder "near-protocl-neps," to "
  - Change "Description" field to "Board Description" and move the field after "Board Title" (should appear second)
14. **Configuring Board Tab:** Change column max text to 10 and only allow up to 10 columns
15. **Configuring Board Tab:** Can we update the column fields as follows:
  - Change "Title" to "Column Title" and the placeholder "Review" to "Enter column title."
  - Change Description placeholder "NEPS that need a review by Subject Matter Experts." to "Enter a brief description of the new column."
  -  Change "Tag" to "Enter a single tag to show posts in this column" and add placeholder "Tag-Name"
16. **Configuring Board Tab:** Do we have the ability to customize which fields to show/hide in the Kanban cards (e.g. post title, post author, post type, post labels, # replies)? It was part of the AC but we can move to a separate ticket if this is not easily done.

carina-akaia:
@ori-near

1. Updated
2. Fixed
3. It happens due to bad VM optimization for React rendering behavior. There __might be__ workarounds for __some__ of the use cases, but it requires some time to figure it out. I'd suggest creating a separate ticket for that.
4. We can't control the tx confirmation window as it is a part of gateway.
5. Same as 4.
6. Addressed in #172
7. This will happen every time a user creates their second and consequent communities, because at the moment we only show **all** communities that have the user in the admin list, without sorting inside this subset. I have a solution in mind which would require some changes on the contract side, so let's also create a separate ticket for this. For now, I can remove this behavior and revers the order in which communities are being rendered, but it's still **not guaranteed** that the order will always be the same. Are you sure we should use this approach?
8. [The design is being discussed]
9. Same as 2.
10. Fixed
11. Done
12. Done
13. Done
14. Yes, let's create a separate ticket.

ori-near:
Hi @carina-akaia – thanks so much for quickly addressing these. The board experience looks awesome!! I will create separate tickets for the issues we've mentioned. I also created a new list with things I've noticed during my last testing:

## Updated list:
1. **Create Communities:** Regarding 7 above, would like to keep the same ordering behavior that exists today if possible, where the newest community shows last. We can revisit this later.
2. **Configuring Board Tab**:  Is it possible to also change the board description placeholder from "Latest NEAR Enhancement Proposals by status." to "Enter board description."
3. **Configuring Board Tab**: Is it possible to reduce the margins slightly between the column title and description? As well as between cards? See [GitHub board](https://github.com/orgs/near/projects/60/views/1) for reference. Screenshot 2023-08-17 at 1 29 35 PM

4.  **Configuring Board Tab**: Can we remove the Form/JSON toggle? I don't think it provides any value right now.
5.  **Configuring Board Tab**: After I filled out the Kanban board configuration form fields and added one column, I clicked Save. Then I got redirected to the initial Board page saying that "This community doesn't have a kanban board." After I refreshed, I was able to see the new board. Ideally the user will see the refreshed new board after clicking save.  Screenshot 2023-08-17 at 1 37 59 PM
6.  **Configuring Board Tab**: Can we add the ability to delete the board (and reset back to the "This community doesn't have a kanban board"? If we need to, we can address this in a separate ticket. But if it's easy to implement that would be nice.

carina-akaia:
1. Done, but the'll hopefully show first
2. Done
3. wip
4. wip
5. wip
6. wip

ori-near:
Hi @carina-akaia – thanks for fixing most of the issues. Looks good! Thanks so much.

PR #232, opened 2023-08-08
feature: introduces better markdown functionality for posts #221

elliotBraem:
Makes changes to post form to feature a more powerful markdown editor

_Resolves https://github.com/near/neardevhub-widgets/issues/221_

PR #255, opened 2023-09-09
SimpleMDE Markdown Editor for posts

### LINKED ISSUE #256: SimpleMDE Editor crashing
**Describe the bug**
DevHub has been crashing or consistently reloading. 

According to profiler, the majority of time is spent in the new simpleMDE editor. This is due to the legacy editing implementation that renders the editors upfront, and just makes it visible/hidden when user clicks on Reply -> Comment.

**Additional context**
The SimpleMDE commit has been reverted. A fix has been underway

elliotBraem:
Resolves #256 

Also removes spell checking at @ori-near's request.

### DIFF
--- a/src/gigs-board/components/molecule/markdown-viewer.jsx
+++ b/src/gigs-board/components/molecule/markdown-viewer.jsx
@@ -1,0 +1,54 @@
+const Wrapper = styled.div`
+  p {
+    white-space: pre-line; // This ensures text breaks to new line
+
+    span {
+      white-space: normal; // and this ensures profile links look normal
+    }
+  }
+
+  blockquote {
+    margin: 1em 0;
+    padding-left: 1.5em;
+    border-left: 4px solid #ccc;
+    color: #666;
+    font-style: italic;
+  }
+  pre {
+    background-color: #f4f4f4;
+    border: 1px solid #ddd;
+    border-radius: 4px;
+    padding: 1em;
+    overflow-x: auto;
+    font-family: "Courier New", Courier, monospace;
+  }
+`;
+
+const Embedded = styled.span`
+  white-space: normal;
+
+  p {
+    white-space: normal;
+  }
+`;
+
+const renderMention =
+  props.renderMention ??
+  ((accountId) => (
+    <span key={accountId} className="d-inline-flex" style={{ fontWeight: 500 }}>
+      <Widget
+        src="neardevgov.near/widget/ProfileLine"
+        props={{
+          accountId: accountId.toLowerCase(),
+          hideAccountId: true,
+          tooltip: true,
+        }}
+      />
+    </span>
+  ));
+
+return (
+  <Wrapper>
+    <Markdown text={props.text} onMention={renderMention} />
+  </Wrapper>
+);
--- a/src/gigs-board/components/molecule/markdown-viewer.jsx
+++ b/src/gigs-board/components/molecule/markdown-viewer.jsx
@@ -13,7 +13,9 @@ const Wrapper = styled.div`
     border-left: 4px solid #ccc;
     color: #666;
     font-style: italic;
+    font-size: inherit;
   }
+
   pre {
     background-color: #f4f4f4;
     border: 1px solid #ddd;
--- a/src/gigs-board/entity/post/Post.jsx
+++ b/src/gigs-board/entity/post/Post.jsx
@@ -500,8 +500,8 @@ const EditorWidget = (postType) => {
 const isDraft =
   (draftState?.parent_post_id === postId &&
     draftState?.postType === state.postType) ||
-  (draftState?.edit_post_id == postId &&
-    draftState?.postType == state.postType);
+  (draftState?.edit_post_id === postId &&
+    draftState?.postType === state.postType);
 
 function Editor() {
   return (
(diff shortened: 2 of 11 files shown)


elliotBraem:
This is a solution for #252 and an extension of #232.

You can preview it [here](https://near.org/devhub.efiz.near/widget/gigs-board.pages.Feed?nearDevGovGigsContractAccountId=devgovgigs.near&nearDevGovGigsWidgetsAccountId=devhub.efiz.near).

src/gigs-board/components/molecule/markdown-viewer.jsx
```diff
@@ -0,0 +1,54 @@
...
+    font-style: italic;
+  }
+  pre {
+    background-color: #f4f4f4;
+    border: 1px solid #ddd;
+    border-radius: 4px;
+    padding: 1em;
+    overflow-x: auto;
+    font-family: "Courier New", Courier, monospace;
+  }
```
frol, line 24: I would consider to contribute these styles to the [Social VM Markdown component](https://github.com/NearSocial/VM/blob/master/src/lib/components/Markdown.js) if you will have time for that.

frol (changes requested):
The tabs are not crashing anymore!

Yet, autocomplete does not work for me now. Is it a known limitation?

src/gigs-board/components/molecule/markdown-viewer.jsx
```diff
@@ -0,0 +1,54 @@
...
+    }
+  }
+
+  blockquote {
+    margin: 1em 0;
+    padding-left: 1.5em;
+    border-left: 4px solid #ccc;
+    color: #666;
+    font-style: italic;
+  }
```
frol, line 16: Can we also make the font size the same as a regular text?

Currently, blockquote text is bigger than the regular text:

image

src/gigs-board/components/molecule/markdown-viewer.jsx
elliotBraem: Good suggestion, but I don't think the VM should impose styles. The Markdown component is only meant to resolve the correct element tags. Developers can fork their own wrapper component w/ styles.

src/gigs-board/entity/post/Post.jsx
```diff
@@ -499,22 +497,54 @@ const EditorWidget = (postType) => {
...
-    {EditorWidget("Sponsorship")}
-    {CreatorWidget("Github")}
-    {EditorWidget("Github")}
-  </div>
-);
+const isDraft =
+  (draftState?.parent_post_id === postId &&
+    draftState?.postType === state.postType) ||
+  (draftState?.edit_post_id == postId &&
+    draftState?.postType == state.postType);
```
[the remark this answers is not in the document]
elliotBraem: Good catch

src/gigs-board/components/molecule/markdown-viewer.jsx
elliotBraem: Here you go!

Screenshot 2023-09-10 at 10 23 51 AM

ailisp:
Is ther preview in https://near.org/devhub.efiz.near/widget/gigs-board.pages.Feed?nearDevGovGigsContractAccountId=devgovgigs.near&nearDevGovGigsWidgetsAccountId=devhub.efiz.near ?

Font still looks bigger in quote:
image

elliotBraem:
@ailisp I had a different story in preview for that. Now that link has the version from this PR with the Quote in same size.

ailisp:
> If you try it on a word in the middle of the text area, it will not pick up on it.

I can confirm this is also the behavior as production and putting the auto complete at the end of text works for this preview. Since @frol also doesn't observe the crashing issue, I think this PR is ready to merge. Frol, do you have other concerns on this PR or approve it?

src/gigs-board/components/molecule/markdown-viewer.jsx
frol: Well, maybe it shouldn’t, but it already does it, so I would at least make the default styling reasonable. It is not a blocker for this PR at all

frol (approved):
> keep in mind that in order for auto complete to work, it needs to be the last characters of text field

Oh, I see, I often type usernames in the middle of the text, so maybe that is why it didn’t work in my previous test. It works as you described, and I think we can work on improvements in a separate PR later, so I approve the changes.

src/gigs-board/components/molecule/markdown-viewer.jsx
elliotBraem: I think you may be thinking about [SocialMarkdown](https://near.social/mob.near/widget/WidgetSource?src=near/widget/SocialMarkdown) widget, which is the wrapper for post previews on near.org. Idk who manages that widget, but they can absolutely put these styles in there (or I can make a PR, but I can't find a public repo for these). The VM holds no CSS and does not apply any stylings, [React-markdown](https://github.com/remarkjs/react-markdown) only renders the markdown -- which brings up a good point -- in documentation, we should ensure a wrapper is utilized with the Markdown primitive.

DevHub didn't have a wrapper widget before, so it was just getting styles from the gateway -- maybe there could be a PR to make the default styling of the discovery-gateway reasonable... but I also feel weary against gateway-imposed stylings...

PR #261, opened 2023-09-15
fix: removed moderators from team page

Tguntenaar:
_Resolves #260_

Preview:
Screenshot 2023-09-14 at 23 24 04


ailisp:
Above part:
```
const root_members = DevHub.get_root_members() ?? null;

if (communityData === null || root_members === null) {
  return <div>Loading...</div>;
}

const moderators = (root_members ?? {})?.["team:moderators"]?.children;
```
root_members and moderators should also be removed to avoid an unnecessary network request

ori-near:
Hi @Tguntenaar – do you need anything to complete this?

PR #264, opened 2023-09-19
Feature/262 solution empty sponsor field

Tguntenaar:
_Resolves #262_

[Preview](https://near.social/previewthomas.near/widget/gigs-board.pages.Feed)

When a user either creates a solution post or replies with a solution and selects 'seeking funding,' the submit button will become disabled if they haven't specified an amount. The requested sponsor field is optional and thus doesn't disable the button when it is left empty. 

Question: Should the requested sponsor field even be optional?

### DIFF
--- a/src/gigs-board/entity/post/PostEditor.jsx
+++ b/src/gigs-board/entity/post/PostEditor.jsx
@@ -161,7 +161,8 @@ const onSubmit = () => {
         state.description,
         state.amount,
         state.token,
-        state.supervisor
+        state.supervisor,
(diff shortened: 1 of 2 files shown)


ori-near:
> Question: Should the requested sponsor field even be optional?

Hey Thomas – yes. Because some people might offer a solution for the good of the community without requiring a compensation. Or offer a pre-existing solution, etc.

ailisp (approved):
Looks great! Now it works well with empty solution, and sponsor field is still optional to fill

PR #284, opened 2023-09-29
Test framework

### LINKED ISSUE #276: Automatic testing

Currently when developing new features or fixing existing, there is no way to be assured that I haven't broken anything existing. As a newly onboarded developer to the project, even experienced, I find it risky to touch any of the DevHub widget code in fear of breaking something.

Also when developing, I always have to do my testing iterations manually in a browser. A development process always involves many roundtrips of tweaking and testing, and not having the possibility to start an automatic test, and instead having to do it manually slows down development velocity.

I would like to be able to develop test driven. When writing new code I would like it to be possible to run a test to see how it behaves. Also I would be able to verify that my changes have not broken any existing features. I would like to be able to run tests both locally in my development environment, and also when pushing it to github. And I would also like to have a test pipeline on github that prevents anything from being merged if the tests are not passed.

**Describe alternatives you've considered**
Testing BOS widgets requires a browser, and so it makes sense to run the tests in a browser. It is also important that the test framework has a small footprint, and runs your tests fast, which is essential for test driven development. 

Cypress is a complete package, with test runner, API and everything, while there are lighter alternatives like the older Karma and the newer PlayWright. Both can achieve the same as Cypress, but the communities around these are smaller. Still both Karma and PlayWright are lighter and faster, and gives more flexibility in choosing test runner ( Jest, mocha, Jasmine ), and also supporting multiple browsers. Another thing is also if the test spans over multiple pages / domains where Playwright might do a better job.

When it comes to interaction with the blockchain, as a start read-only tests using NEAR mainnet could be implemented, this should cover many uses cases for testing UI features like listings, appearance of forms, and it should also even be possible to intercept calls to make transaction without actually performing them.

**Additional context**

petersalomonsen:
Setup of [Playwright](https://playwright.dev) running tests via https://near.org with a bos-loader configuration to load DebHub widgets served from localhost.

See the changes to CONTRIBUTING.md as well as the tests inside the playwright-tests folder containing a test for verifying the bos-loader setup, and an example test for verifying that a post in the feed contains the post history button.

also added gitpod.yml that contains a setup for [gitpod.io](https://gitpod.io), which I use for developing.

fixes #276


frol:
@petersalomonsen Great job! This plan looks good to me!

frol:
My only concern so far is that CI takes 50 seconds to install dependencies, and 10 seconds to run a single test. I hope it will scale well, and we won't see sequential tests each taking 10 seconds to run.

petersalomonsen:
@frol Playwright supports running tests in parallel ( https://playwright.dev/docs/test-parallel ) , and I've now enabled this for CI as you can see from the latest test run.

The main reason for each test taking about 10 secs is because we are waiting for near.org to load. We can add an issue for improvement later to have a local vm/gateway and even also a lightweight smart contract simulator (like proposed earlier) to mock the RPC, or even just mock the RPC. This should speed up the execution time for a single test significantly. Still though, since we depend on other external widgets, (like the post history for example), these also need to load, so we could consider preloading these.

Loading the dependencies involves loading all the browsers. We might not want to test in all these, so we might reduce time by loading fewer. Still I think it's a good thing that we download the latest browser versions, but we can maybe find a way to cache these for some days at least.

All these are improvements that we can add followup issues for.

I think this test framework MVP is more or less ready now, so unless you think there is something else, I'll move this from draft to ready for review.

frol (approved):

frol:
Let’s get it merged and finally get into a habit of having tests written to speed up our development!

Issue #248, opened 2023-08-28
Author-based searched is outdated


nearmax:
**Describe the bug**
When I search for posts from `nearmax.near` I only see very old posts:
![image](https://github.com/near/neardevhub-widgets/assets/35039879/1f766a9e-55b1-4fdb-8656-0bb58bfb54fe)
and it does not show more recent posts that I made:
https://near.social/devgovgigs.near/widget/gigs-board.pages.Post?id=1034

**To Reproduce**
1. Go to https://neardevhub.org/
2. Search for posts made by `nearmax.near`;
3. Observe that https://near.social/devgovgigs.near/widget/gigs-board.pages.Post?id=1034 is not coming up

**Expected behavior**
All posts from `nearmax.near` are shown.

ailisp:
@Rishabhraghwendra18 Thanks for looking into this. This is fixed in #298

Issue #115, opened 2023-04-26
Server side search, filter and sort


ailisp:
**Problems**

Currently, the search, filter by author or label, sorting by newest or latest are happens in smart contract and frontend. Then the result, such as search, filter and sorting is combined together in the frontend. There is a couple of problems with this approach:
- It performs bad and doesn't scale when many more posts are added in future. Even now we cannot search non top level posts for performance reason.
- The full text fuzzy search isn't robust. It can produce confusing result because goes far for trying fuzzing the search result.
- It is implemented in a hacky way. The code to do combination in frontend is difficult to read and maintain. It is hard to extend to more kinds of filters and sorting in future.

**User Story**
As a user of DevHub, I want the search, filter and sort experience to be smooth and reliable. They should also work together. As a maintainer of DevHub, I want this part of code is easy to maintain and can be extend for future needs.

**Acceptance Criteria**

- Implement the indexer from data team's QueryAPI Dashboard, as demoed here: https://www.loom.com/share/77ab9a4acbf34e1aa65961dc70459a22 to avoid maintain an indexer server infra.
- The indexer can do full text search posts. If there is a precise match, the result should be returned. Otherwise, the fuzzy search result is returned.
- The indexer can filter the post by author and label
- The indexer can sort the post by newest and hottest
- The indexer can combine above results.
- Remove frontend combination resullt logics from widgets
- Remove unnecessary smart contract side filter logics from contract

**Technical Details**

1. [The most recent version of QueryAPI widget](https://near.org/dataplatform.near/widget/QueryApi.App)
2. Search DevHub contract stored Posts, which technically it's from "add_post", "edit_post" transaction content.
3. Query API may or may not work with historical blocks but only new blocks. If not, it means only posts after the indexer launched will be processed and indexed and we'll need to do a one-time processing for existing posts either from historical blocks, or a dump of contract state.

** Spike breakdown **
- [x] Create database schema to store necessary info extract from transactions
- [x] Create indexingLogic.js to populate data and store in database
- [ ] Create graphql queries to read (search & filter) from the database
- [ ] Call graphql queries from frontend and replace current search & filter logics

near-akaia-root:
There's probably a blocker, since indexer deployment requires an access key for `queryapi.dataplatform.near. I actually think that this feature still can be properly implemented entirely on frontend, but it would require at least 2 weeks to achieve a good architectural approach in order to make it well-maintainable.

For now, I'm switching to a P2 task until further updates take place.

[...]

Issue #106, opened 2023-04-21
Search for posts not finding posts with specific keywords


ori-near:
**❌ Dependency**
This is dependent on https://github.com/near/neardevhub-widgets/issues/115 getting completed first.

**Problem**
Although we recently introduced search capability in #66, the current design implementation has some [limitations](https://github.com/near/neardevhub-widgets/pull/85#issuecomment-1500202212) that lead to a poor user experience. To name a few examples:

- The search only works for top level parent posts, meaning that users are unable to search for child level posts
- The fuzzy search sometimes omits explicit terms
- The search experience is slow and results in freezes such as when searching for long text queries

**User Story**
As a DevHub user, I want to search for any posts that mention a specific topic or keyword even if I don't know the exact phrasing or spelling.

**Acceptance Criteria**

- The search works for any posts (top level or child)
- The search function uses fuzzy matching to return posts that match the search term, including partial matches and misspelling. The search should always prioritize the explicit terms first
- The search results should load quickly
- The search function displays results as they are found, allowing users to see posts matching the search in real-time

---

### Example of searching for terms in child posts

**To reproduce:**
Here's an example of child terms that get omitted:

1. Log into the [DevHub](https://near.social/#/devgovgigs.near/widget/gigs-board.pages.Feed) as theori.near
2. Filter by `zero-knowledge` labels and the `devgovgigs.near` author
3. Search for "typically"
4. Search output returns: "No posts matches search"

**Expected behavior:**
Search output should reveal 2 posts that match these conditions ([post 1](https://near.social/#/devgovgigs.near/widget/gigs-board.pages.Post?id=372), [post 2](https://near.social/#/devgovgigs.near/widget/gigs-board.pages.Post?id=373))

**Desktop:**
OS: OS
Browser Chrome
Version: 112.0.5615.49

**Screenshots:**
![Screenshot 2023-04-20 at 6 10 34 PM](https://user-images.githubusercontent.com/110252255/233517531-1774cbe6-f825-4685-8a2b-c94a9d1e897b.png)

https://user-images.githubusercontent.com/110252255/233517605-6e96a58f-371f-4a32-a223-e5378aab7523.mp4

frol:
[...]

PR #298, opened 2023-10-13
Implement server side search

### LINKED ISSUE #292: Search for post field doesn't yield any results
**Describe the bug**
Currently I am unable to search for any posts using the search field. This makes it impossible for me to find posts.

**To Reproduce**
1. Go to DevHub's homepage
2. Type any key word in the "Search Posts" field
3. See hovering circle inside the search posts (e.g. loading button) without any search outputs

**Expected behavior**
I should see the posts that contain the search term/

**Screenshots**
Screenshot 2023-10-10 at 4 22 14 PM

 - OS: iOS
 - Browser: Safari
 - Version: Latest

ailisp:
Implement https://github.com/near/neardevhub-widgets/issues/115. Fix #292 #248 #106

The search code is in QueryAPI, code is public here: https://near.org/dataplatform.near/widget/QueryApi.App?selectedIndexerPath=bo.near/devhub_v17&view=editor-window can be forked and updated if there's update in post add / edit format.

### Preview

https://near.social/bo.near/widget/gigs-board.pages.Feed?nearDevGovGigsContractAccountId=devgovgigs.near

### Testing
search one of author, tag or content; search author + tag, author + content, tag + content; search author + tag + content. All testing produce correct result.

### Notes

- A design for search result pagination is needed to have that pagination on the UI. It currently hard coded to display top 100 search result, to make server response faster enough. In practice, it's not a problem because user is very unlikely to browse more than 100 search result instead of trying to refine the search. So I think pagination isn't a priority. Pagination on server is  implemented. 
- server has more capability than frontend shows, for example filter by post category, sort with some criteria, search top level posts only or search posts & comments (current behavior), title, history versions, sponsor, editors, etc. These can be easily enabled if design. We may need to think of a new design of the search panel.
- A "Search" button is added to trigger search to avoid the previous stuck when typing experience. Also due to [a limitation of VM ](https://github.com/NearSocial/VM/issues/139) we cannot trigger search with `<Enter>` key. To make the search experience consistent, **now search by author or tag also doesn't trigger until you click "Search" button**
- Same as the behavior before, you also must click the auto-completion item, to search by author or tag.
image
doesn't set author to `frol.near`
image
does set author and now click search can search author by `frol.near`

This is suboptimal and hopefully we can find some workaround to change this `<Typeahead>` behavior

### DIFF
--- a/playwright-tests/search.spec.js
+++ b/playwright-tests/search.spec.js
@@ -1,0 +1,20 @@
+import { test } from "@playwright/test";
+
+test("should show post history for posts in the feed", async ({ page }) => {
+  await page.goto("/devgovgigs.near/widget/gigs-board.pages.Feed");
+
+  // Fill the search by content by to
+  const searchInputSelector = 'input.form-control[type="search"]';
+  let searchInput = await page.waitForSelector(searchInputSelector, {
+    state: "visible",
+  });
+  await searchInput.fill("zero knowledge");
+
+  const searchButton = await page.getByRole("button", { name: "Search" });
+  await searchButton.click();
+
+  await page.waitForSelector('span:has-text("zero knowledge")', {
+    state: "visible",
+    timeout: 10000,
+  });
+});
--- a/src/gigs-board/entity/post/List.jsx
+++ b/src/gigs-board/entity/post/List.jsx
@@ -126,9 +126,9 @@ function defaultRenderItem(postId, additionalProps) {
 const renderItem = props.renderItem ?? defaultRenderItem;
 
 const cachedRenderItem = (item, i) => {
-  if (props.searchResult && props.searchResult.keywords[item]) {
+  if (props.searchResult && props.searchResult.keywords) {
     return renderItem(item, {
-      searchKeywords: props.searchResult.keywords[item],
+      searchKeywords: props.searchResult.keywords,
     });
   }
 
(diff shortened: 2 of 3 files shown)


frol:
@ailisp Does the search work for the post titles? I cannot find the fellowship program announcement post with either of the following queries: "devdao fellowship", "fellowship program"

ailisp:
@frol it does not. it's support on server but the UI is search by content only.

Search by title need some design, because for a title: "Sponsorship: XXX" the title is actually only "XXX" and it could be confused to not see if they search for "Sponsorship" and didn't get it. It's also support on server side to search by post category, e.g. "Sponsorship". How to present these search criteria and make them all intuitive, no confusion requires more think about the search panel. So I keep search panel minimally changed for now.

petersalomonsen:
How do we ensure that the query is updated if someone changes the post format. Should we add a test for the query? Also I find the reason for #292 to be that some posts does not deserialize anymore. This will still be a problem, and not fixed by this PR. Maybe we should rephrase #292 since you are now changing to another way of searching?

petersalomonsen (approved):
I think this looks very good! If possible to add a playwright test ( checking that the query works ), then that would be great. But I've tested the preview and I think it works fine. 

Let me know what to do with #292 . The bug in the contract is still there, and I can still look into it if we want to.

petersalomonsen:
> It's a great investigation, however it seems like not the bug that @ori-near reported. Ori said search any term fail to find result, not just post 146.

The current search functionality use `get_posts` which currently does not work because of some posts failing to deserialize ( 146 among others ). So that's why it does not find any results, but anyway that problem of getting results will be solved by your PR here.

ailisp:
I added one search test with @petersalomonsen's testing framework and will merge this. Will talk to @ori-near on more search criteria: filter by post category, sort with some criteria, search top level posts only or search posts & comments (current behavior), title, history versions, sponsor, editors, etc. They are easy to add but we need to figure out what we need.

Issue #314, opened 2023-10-18
Community pages showing all posts


PiVortex:
**Describe the bug**
Under the community pages, under activity, all posts are showing,

**To Reproduce**
1. Go to any community pages [for example](https://near.social/devgovgigs.near/widget/gigs-board.pages.community.activity?handle=component-libraries)
2. Go to activity, look in the list
3. See all posts

**Expected behaviour**
Only posts made under the community or tagged with the community should be shown here 

**Screenshots**
![image](https://github.com/near/neardevhub-widgets/assets/124691791/9635e8fe-ed65-4640-bba7-cff4f4f08730)
Component libraries but near campus is showing

**Desktop (please complete the following information):**
OS: Windows 11
Browser: Chrome

ori-near:
Thanks @PiVortex for reporting this. Confirming that I reproduced this.

PR #317, opened 2023-10-19
Improve search ux

ailisp:
- remove the search button, instead press enter to search
- after click an author or label instantly trigger the search
- keep screen not scrolling when doing a search

Preview: https://near.social/bo.near/widget/gigs-board.pages.Feed?nearDevGovGigsContractAccountId=devgovgigs.near

### DIFF
--- a/playwright-tests/search.spec.js
+++ b/playwright-tests/search.spec.js
@@ -9,9 +9,7 @@ test("should show post history for posts in the feed", async ({ page }) => {
     state: "visible",
   });
   await searchInput.fill("zero knowledge");
-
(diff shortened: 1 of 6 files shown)


frol (approved):
I feel the "Search" button might still make sense from the UX point of view as it might be not clear to all the users that they need to press enter to trigger search, but I am not going to block this PR.

ailisp:
@frol Discussed your idea with Ori, we decide to move the search icon to the right and make it clickable as the search button. And a few other UI/UX tweaks. I'll merge this to work on #324 and #314 and revisit these UI/UX things after them

Issue #226, opened 2023-08-03
Filtering by label provided in the URL is not respected


frol:
**Describe the bug**

Using label chips on the posts should apply filtering by that label, but it simply shows an unfiltered list of posts:

https://near.org/devgovgigs.near/widget/gigs-board.pages.Feed?label=near-sdk-js

This link should only show the posts labeled with `near-sdk-js`, but it does not.

**To Reproduce**
1. Go to the feed
2. Click on any label chip on any post to get navigated to `?label=xxx` page
3. The list of posts is unfiltered

**Expected behavior**

The list should be filtered as if you entered the label in the search field. Also, the search field should be populated with the value provided in the URL.

**Screenshots**

<img width="959" alt="image" src="https://github.com/near/neardevhub-widgets/assets/304265/f0a87e06-d24b-479e-a14a-35b6f5cb7d66">

**Desktop:**
 - OS: any
 - Browser any
 - Version latest

**Smartphone:**
 - Device: any
 - OS: any
 - Browser: any
 - Version any

jaswinder6991:
I can take this. Can you please assign it to me.

ori-near:
Hi @jaswinder6991 thanks so much for expressing interest 🔥. We appreciate your willingness to contribute!

We have assigned this ticket to you and would like to kindly request that you submit a pull request with your solution within one week. We moved this ticket to In Progress to indicate that you are working on it. If you are no longer able to work on this issue, please let us know so we can reopen it for another developer to take on. Once again, thanks so much and we look forward to your contribution.

jaswinder6991:
Thanks @ori-near 
Here is what I found, the label prop does not really work on the feed widget, you instead have to do ```tag``` , so the working URL should be.
https://near.org/devgovgigs.near/widget/gigs-board.pages.Feed?tag=near-sdk-js

As I went down the widget rabbit hole, I found out the tag widget which is this:
https://near.org/devgovgigs.near/widget/gigs-board.components.atom.tag 

Here is the problem in this widget, current code:
```
function href(widgetName, linkProps) {
  linkProps = { ...linkProps };

  if (props.nearDevGovGigsContractAccountId) {
    linkProps.nearDevGovGigsContractAccountId =
      props.nearDevGovGigsContractAccountId;
  }

  if (props.nearDevGovGigsWidgetsAccountId) {
    linkProps.nearDevGovGigsWidgetsAccountId =
      props.nearDevGovGigsWidgetsAccountId;
  }

[...]

PR #327, opened 2023-10-20
a few search and sort enhancements

### LINKED ISSUE #314: Community pages showing all posts

### LINKED ISSUE #324: Display all posts including replies
It is hard to see of activity happening on DevHub now that there is no option to see all replies.

A button that shows all replies.

**Describe alternatives you've considered**
I couldn't come up with a better alternative.

ailisp:
Fix #324, Fix #314 
Also make search instantly trigger when typing, instead of pressing enter to search.

Preview: https://near.social/bo.near/widget/gigs-board.pages.Feed?nearDevGovGigsContractAccountId=devgovgigs.near

Note:
- Search by default seaching only posts, not comments
- Select "Sort -> All replies" it will list latest comments and posts
- After "Sort -> All replies" then search will search all comments and posts
- Click on a community now correctly list posts of that community


frol (changes requested):
> After "Sort -> All replies" then search will search all comments and posts

This is not really intuitive since it is not "sorting". I guess you may want to rename it to "filter" and default would be "Top-level posts".

There is also a bug: the state of this dropdown is not displayed after selection.

frol:
Another bug related to #314 - tags filtering does not work when you click on a tag of a post (see #226) it was previously resolved, but it is an issue again.

ailisp:
@frol 

> This is not really intuitive since it is not "sorting". I guess you may want to rename it to "filter" and default would be "Top-level posts".

I have the same feeling here, your proposed way is more intuitive to me, but this is trying to match original design: 
image

(Sorting by hottest is WIP in indexer side). Let's come back to consider it after getting it matches to the design.

I'll fix these ones:

> There is also a bug: the state of this dropdown is not displayed after selection. 
> Another bug related to https://github.com/near/neardevhub-widgets/issues/314 - tags filtering does not work when you click on a tag of a post (see https://github.com/near/neardevhub-widgets/issues/226) it was previously resolved, but it is an issue again.

src/gigs-board/feature/post-search/panel.jsx
```diff
@@ -63,6 +67,8 @@ const buttonStyle = {
   color: "#f3f3f3",
 };
 
+console.log(state.tag);
```
frol, line 70: Remove the debug logs

frol (approved):
All the reported issues are addressed and this PR looks good to me

PR #335, opened 2023-10-23
DevHub Revamp

elliotBraem:
View the preview [here](https://test.near.org/devhub-dev.testnet/widget/app).


yarn.lock
frol, line 1: Let's not use yarn. There is no reason for it.

CONTRIBUTING.md
```diff
@@ -107,7 +107,7 @@ NOTE: If you have [DevHub contract](https://github.com/near/neardevhub-contract)
 
 #### Deploy for Production
 
-There is GitHub Actions automation that deploys all the widgets to [`devgovgigs.near` account](https://near.social/#/mob.near/widget/MyPage?accountId=devgovgigs.near) on mainnet on every push to the main branch.
+There is GitHub Actions automation that deploys all the widgets to [`devgovgigs.near` account](https://near.social/#/${REPL_MOB}/widget/MyPage?accountId=devgovgigs.near) on mainnet on every push to the main branch.
```
frol, line 110: I think it was a change caused by a replace-all, but it should not be done here. Let's revert this change

scripts/deploy.sh
```diff
@@ -0,0 +1,142 @@
...
+  sudo apt-get update
+  sudo apt-get install -y jq
+fi
+
+# Check if sed is installed
+if ! command_exists sed; then
+  echo "sed is not installed. Installing..."
+  sudo apt-get update
+  sudo apt-get install -y sed
+fi
```
frol, line 20: Woha. This is totally out of scope for a script like this.

Have you considered using JS for a script? ([`/module/include.js`](https://github.com/near/neardevhub-widgets/blob/main/module/include.js) historically managed the replacements)

src/app.jsx
frol, line 1: I am yet to see the benefits of this design as it clearly creates another layer of indirection => slower load, lost ability to view the source code of the page using the BOS "Develop" menu (all the pages will be "app" pages), and you also need to remember to add every new page here.

frol (changes requested):
This is a monster of a PR. Is there a chance to split the changes into smaller bits?

scripts/deploy.sh
frol: @elliotBraem You may take inspiration from my demo TypeScript starter where I also generate a build folder on the fly: https://github.com/frol/bos-component-ts-starter/blob/main/package.json#L25 (see `node ./build.js` at the end, and see the [`build.js` file](https://github.com/frol/bos-component-ts-starter/blob/main/build.js))

src/app.jsx
elliotBraem: How do you think it is slower? It's loading much faster for me... let me know what is slower and I can investigate why. 

There are two main reasons for this decision:

1) We are getting a new homepage and this involves a new font and layout; we want theme to be at the root and pass to all the children. How else would you propose to solve this?
2) We want to use <Link /> and route around a common route in order to reduce reloads. You'll notice in the current application, entire pages will rerender on navigation. Now, the Layout stays consistent and the changes are contained to what the client determines it needs to rerender. It's just react-router-dom concepts, but we don't have an "App.jsx" at our core, so we need to make our own "App" root. You'll see similar structures in other more complex BOS apps, such as [Astra ++](https://near.social/astraplusplus.ndctools.near/widget/index) and [Horizon](https://near.social/#/nearhorizon.near/widget/Index).

Yes, you are unable to view the source code of the page from "Develop" menu, but what would you suggest as an alternative? This is inherent to any app on BOS that isn't a standalone widget. I would argue that this solution provides a clearer path to understanding and finding the dependent child because of this separation.

elliotBraem:
> This is a monster of a PR. Is there a chance to split the changes into smaller bits?

@frol 

Haha. I'll remove the replacements that happens in the gigs-board namespace and the one you mentioned -- there aren't supposed to be any changes to the current app, it's supposed to all be contained to devhub. 

Then I can also break it down into these:

1) New root and refactor, no community addon framework
2) Community addon framework and no plugins
3) Plugins

I've made the first of these three PRs, you can find it here: #341

scripts/deploy.sh
elliotBraem: I've removed the deploy script from all of the plugins. I think this would make a great ticket

elliotBraem:
I've repurposed this PR for the devhub revamp.

This branch has been reset to the code in #341; a refactor of the devhub app with a new root at devgovgigs.near/widget/app

ailisp (approved):

PR #341, opened 2023-10-24
App refactor and new root

elliotBraem:
_Resolves #294_

This refactors the existing app into having a new root for unblocking the new homepage stories and enabling the community addon framework.

See the demo [here](https://near.social/devhub.efiz.near/widget/app).

This PR introduces no changes to the current devhub application.

There are two main reasons for establishing this root and directing pages from /app:

1) We are getting a new homepage and this involves a new font and layout; we want theme to be at the root and pass to all the children. 
2) We want to use <Link /> and route around a common route in order to reduce reloads. You'll notice in the current application, entire pages will rerender on navigation. Now, the Layout stays consistent and the changes are contained to what the client determines it needs to rerender. It's just react-router-dom concepts, but we don't have an "App.jsx" at our core, so we need to make our own "App" root. You'll see similar structures in other more complex BOS apps, such as [Astra ++](https://near.social/astraplusplus.ndctools.near/widget/index) and [Horizon](https://near.social/#/nearhorizon.near/widget/Index).


petersalomonsen:
Moving the page part of the URL into the query string might affect SEO. It's also going against "Clean URL" principles ( https://en.wikipedia.org/wiki/Clean_URL ). Are we sure we want such a change. There are ways to obtain having the theme at the root without breaking clean URLs.

elliotBraem:
What would you suggest? I want to be careful to not confuse a BOS app with a typical react app because we do not have the same freedoms, e.g. a clean url that separates like "devgovgigs.near/widget/app/{page}/{id}". This would require changes on the gateway level, because otherwise this full path gets passed as a Widget src. What is your idea for having theme at the root?

This follows the pattern of other larger, complex apps on BOS like [Astra ++](https://near.social/astraplusplus.ndctools.near/widget/index) and [Horizon](https://near.social/#/nearhorizon.near/widget/Index).

scripts/dev.sh
```diff
@@ -1,30 +1,44 @@
...
 
+# Function to check if a command is available
+command_exists() {
+  command -v "$1" >/dev/null 2>&1
+}
+
+# Check if jq is installed
+if ! command_exists jq; then
+  echo "jq is not installed. Installing..."
+  sudo apt-get update
```
carina-akaia, line 11: This will only work on a small portion of OSs, so this code doesn't make sense until we have enough time to cover every case. I would propose to just leave the following link in the docs instead:

https://jqlang.github.io/jq/download/

src/core/adapter/devhub-contract.jsx
carina-akaia, line 1: Are you sure modules should have the .jsx file extension too even if there's no JSX?

carina-akaia (changes requested):

src/core/adapter/devhub-contract.jsx
elliotBraem: [bos loader](https://github.com/near/bos-loader/tree/main/src) looks for components with .jsx, so this would be a bug to file with them

src/devhub/components/molecule/MarkdownViewer.jsx
```diff
@@ -0,0 +1,56 @@
...
+    white-space: normal;
+  }
+`;
+
+const renderMention =
+  props.renderMention ??
+  ((accountId) => (
+    <span key={accountId} className="d-inline-flex" style={{ fontWeight: 500 }}>
+      <Widget
+        src="neardevgov.near/widget/ProfileLine"
```
ailisp, line 42: This file seems not exist

elliotBraem: Where did the error come up? 


Here is the neardevgov.near/widget/ProfileLine widget: https://near.social/neardevgov.near/widget/ProfileLine
And the MarkdownViewer: https://near.social/devhub.efiz.near/widget/devhub.components.molecule.MarkdownViewer

ailisp (approved):
Great work! Overall it looks great and since it runs on a different url in parallel to production, I think it's fine to merge. Some issues I found in this PR:
- It doesn't take some recent changes in main branch, to name a few:
  - the navbar when in a community in production is different to the navbar in the demo:
image

  - the search bar is updated in production than in this version 
  - USDC is added in production recently
- I don't see most types of tabs in communities:
image

elliotBraem:
Awesome, thank you @ailisp for the review! 

Regarding the issues:

- The navbar difference will be resolved with redesign
- I've made an issue to ensure the search bar is up to date: #353 
- I've made an issue to ensure the USDC is migrated #352 
- Most types of tabs do not appear because this is being replaced with the addon framework

Will merge now

Issue #267, opened 2023-09-20
Migrate existing Wiki plugin to new framework


ori-near:
**Problem**
Currently, community admins only have the option to enable up to two Wiki plugins in the community settings page. However, some communities may want to add additional wikis.

**User Story**
As a community admin, I want the ability to decide if to enable or disable multiple Wiki pages, using a consistent way as other plugins.

**Acceptance Criteria**
- Wiki plugin aligns with the new framework (#253). So an admin can enable, configure, or remove the plugin in the community settings page.
- If a community admin enables the Wiki plugin, the Wiki tab shows up on the community page navigation. And the page itself is populated with the Wiki content set in the Wiki plugin
- If an admin enables the Wiki plugin, they can click configure to customize the the tab name as well as the Wiki specific fields, including:
  - Content (must support markdown language, enabled by https://github.com/near/neardevhub-widgets/issues/178)
- Community admin should have a way to preview Wiki from the configure settings. 
- If an admin enables the Wiki plugin, the Wiki content field is required
- There is no disruption to old/existing Wiki plugins
- There is migration option provided for existing communities to transition to the new framework

elliotBraem:
Included in this scope should be creating a "CommunityAddon"; an instance in the devhub contract:

```rust
pub struct CommunityAddOn {
    id: String,
    title: String,
    description: String,
    viewer: String,
    configurator: String,
    icon: String
}
```

Wiki has it's own schemas, but will need a configurator widget created. After this is added, most of the scope should be completed because of the framework. This migration looks minimally invasive.

Issue #253, opened 2023-09-09
Introduce community plugin management framework


petersalomonsen:
# Product User Story

**Problem**
Currently, community admins do not have a consistent and easy way to personalize their community experience. For example, the GitHub and Kanban  components always display, and an admin must manage them within the tab and cannot remove them. 

**User Story**
As an admin, I need a consistent and intuitive method to add, configure, and remove add-ons for my community via the community settings page.

**Acceptance Criteria**
- By default, any newly created communities will not display any of the available add-ons. So only the main feed and teams tab will be enabled.
- Admins have a consistent framework for adding, configuring, and deleting add-ons all through the community settings page.
- Adding: Admins can add add-ons through a "Add add-on" button and select from a dropdown menu of available add-ons. An admin can add multiple instances of the same add-on. By default, the naming of each tab will be the add-on name.
- Configuring: Once a new add-on is added, admins can click edit to customize the add-on’s tab names and specific fields, and click save to apply their changes. 
  - Previewing: Admins should have a way to quickly preview their tab in the settings page. Perhaps we add a toggle here similar to GitHub's "Write/Preview" interface to avoid crowding the interface for more complex plugin configurations?
- Deletion: Admins can delete add-ons by accessing the edit option and selecting delete, removing the respective tab.
- Migration: Existing add-on configurations are preserved, and communities can manually migrate at their convenience without breaking functionality.
- Tab names
   - Limit # of characters in tab to X
   - Limit # of widgets: Only allow up to ~# number of widgets to fit under the navigation

----------

# Old Technical Notes

**Is your feature request related to a problem? Please describe.**
Currently a DevHUB community only allows built-in pages/components like Telegram and Github, limiting the potential for unique functionalities tailored to the needs of each community. Also the current way of adding Telegram and Github is by registering handles/URLs in the central configuration section, which does not give the experience of a "pluggable" component with its own configuration.

**Describe the solution you'd like**
I would like Telegram, Github and BOS widgets in general to be components that can be added from the configuration page, and also have their own configuration sections. A user experience with a button for adding a "widget" like the following screenshot:

<img width="400" alt="image" src="https://github.com/near/neardevhub-widgets/assets/9760441/1a03958c-4102-4789-9b05-55dc317a0d9b">

When clicking "Add widget" you should be able to select from a drop-down menu different alternatives of built-in (like Telegram and Github) as well of community provided widgets: 

[...]

Issue #175, opened 2023-06-19
Migrate existing Telegram plugin to new framework


ailisp:
**Problem**
Currently, community admins do not have an intuitive way to enable and manage the Telegram plugin. When an admin updates the "Telegram handles" field in the About section, that adds the Telegram tab.

**User Story**
As a community admin, I want the ability to decide if to enable or disable the Telegram. I also want to clearly understand where to configure this page in a consistent way.

**Acceptance Criteria**
- Telegram plugin aligns with the new framework (#253). So an admin can enable, configure, or remove the plugin in the community settings page.
- If a community admin enables the Telegram plugin, the Telegram tab shows up on the community page navigation. And the page itself is populated with the Telegram content set in the Telegram plugin section. Not from the About section.
- If an admin enables the Telegram plugin, they can click configure to customize the the tab name as well as the Telegram specific fields, including:
  - Telegram channels. Ideally, admins can add multiple Telegram channels, each in a separate field box. And each field would have a prefix that says: "https://t.me/" to make it clear what part of the Telegram URL the user needs to enter.
- Community admin should have a way to preview GitHub board from the configure settings. 
- If an admin enables the Telegram plugin, the Telegram field(s) are required
- There is no disruption to old/existing Telegram plugins
- There is migration option provided for existing communities to transition to the new framework

ori-near:
@carina-akaia Thanks for creating this ticket. I added very clear testable acceptance criteria to ensure we don't miss anything. Please let me know if you have any questions.

petersalomonsen:
Tested this a bit for myself now, and also set it up in my preview environment ( no changes at this point ).

https://near.social/devgovgigs.petersalomonsen.near/widget/gigs-board.pages.community.activity?handle=special

I see that adding a Telegram handle adds a Telegram tab. I believe that we should ( rather than having github/telegram as default widgets - even with a toggle ) be able to add any widget, so rather have an "add widget" button like shown in the screenshot below.

And when clicking that button, a dropdown to select the widget you want to add. Finally when added you get a configuration section for that widget, and also the option to delete it.

<img width="737" alt="image" src="https://github.com/near/neardevhub-widgets/assets/9760441/1a03958c-4102-4789-9b05-55dc317a0d9b">

ori-near:
[...]

PR #343, opened 2023-10-25
Community Addon Framework, Wiki and Telegram Plugins

elliotBraem:
_Resolves #253 - Introduce community plugin management framework_
_Resolves #267 - Wiki plugin_
_Resolves #175 - Telegram plugin_

[Preview](https://near.org/devhub.efiz.near/widget/app?page=community&handle=devhub-test)

**Acceptance Criteria**

### Community Addon Framework

- [x] Newly created communities do not display any add-ons, only main feed and teams tab.
- [x] Admins have a consistent framework for adding, configuring, and deleting add-ons all through the community settings page.
- [x] Admins can add add-ons through a "Add add-on" button and select from a dropdown menu of available add-ons. An admin can add multiple instances of the same add-on. By default, the naming of each tab is the add-on name.
- [x] Once a new add-on is added, admins can click edit to customize the add-on’s tab names and specific fields, and click save to apply their changes.
--Add-on parameters can be configured from the add-on itself--
--Some small work could be done to allow admins to preview their changes, this is currently not available--
- [x] Admins can delete add-ons by accessing the edit option and selecting delete
- [ ] Existing add-on configurations are preserved
--There is no migration step--
- [x] Tab names are limited to X # of characters
--This has not yet been implemented--
- [x] There is a limit of 7 addons

### Wiki Plugin

- [x] Wiki plugin aligns with the new framework
- [x] If a community admin enables the Wiki plugin, the Wiki tab shows up on the community page navigation. And the page itself is populated with the Wiki content set in the Wiki plugin
- [x] If an admin enables the Wiki plugin, they can click configure to customize the the tab name as well as the Wiki specific fields
- [x] Community admin should have a way to preview Wiki from the configure settings.
- [ ] If an admin enables the Wiki plugin, the Wiki content field is required
--A wiki plugin can be created, and then needs to be configured directly to add content--
- [x] There is no disruption to old/existing Wiki plugins
- [ ] There is migration option provided for existing communities to transition to the new framework
--There is no migration step--

### Telegram Plugin

- [x] Telegram plugin aligns with the new framework
- [x] If a community admin enables the Telegram plugin, the Telegram tab shows up on the community page navigation. And the page itself is populated with the Telegram content set in the Telegram plugin section. Not from the About section.
- [x] If an admin enables the Telegram plugin, they can click configure to customize the the tab name as well as the Telegram specific fields
- [ ] If an admin enables the Telegram plugin, the Telegram field(s) are required
--A telegram plugin can be created, and then needs to be configured directly to add a telegram handle--
- [x] There is no disruption to old/existing Telegram plugins
- [ ] There is migration option provided for existing communities to transition to the new framework
--There is no migration step--


elliotBraem:
Hi @ori-near, this story is ready for your review. It adds the addon framework, telegram, and wiki plugins.

You can test this out [here](https://near.org/devhub.efiz.near/widget/app?page=community&handle=devhub-test). Feel free to add, remove, and configure the plugins in this community.

src/core/adapter/devhub-contract.jsx
```diff
@@ -113,30 +113,30 @@ function getAvailableAddons() {
...
-    },
-    {
-      id: "blog",
-      title: "Blog",
-      description: "Create a blog for your community",
-      view_widget: "${REPL_DEVHUB}/widget/devhub.entity.addon.blog.Viewer",
-      configurator_widget:
-        "${REPL_DEVHUB}/widget/devhub.entity.addon.blog.Configurator",
-    },
+    // {
```
petersalomonsen, line 116: would we like to have commented code like this merged?

elliotBraem: It get's uncommented when the plugins get merged; I can clean it out if we want, but it get's added right back in until #358 gets completed, but this is no blocker

ori-near:
Hi @elliotBraem  – thanks so much for all your hard work on this! Please see my feedback below. I'm categorizing them as follows: 🔴 Must have 🟠 Nice to have 🟡 Maybe later. Please let me know if you have any questions or want to move any of these items to a separate ticket.

## Feedback

🔴 1. **Configure Community:** Is it possible to show the add-on in the configure community page after adding it? Currently, when I add a plugin, click submit, and go through the transaction confirmation screen, I see the profile page again without the add-on:
Screenshot 2023-10-26 at 9 04 40 PM
If I click refresh, I'm able to see it:
Screenshot 2023-10-26 at 9 04 52 PM
🔴 2. **Configure Community:** Change Addons to "Add-Ons". Also add a text underneath that says: "Add or remove custom tabs, which will appear in your community's navigation bar. You can customize them on each page."
🟠 3. **Configure Community:** Can we add Tab Name field to help indicate what that field represents? If that's too complex we can just leverage the message in item 2 above. 
🟡 4. **Configure Community:** Can we add a character limit, similar to the other admin sections?
🟡 5. **Configure Community:** When I add 3 add-ons, the up/down arrows are shaded in different colors. For example, in the first instance the up is light and the down is dark. The second instance both are dark. The last instance the up is dark and the down is light. Can we make them all consistent or is there a reason for this?
🟡 6. **Configure Community:** Is there a reason why the Addons doesn't follow the same "edit" logic as the other sections (e.g. Access Control, About)? Right now every time I load the page the form fields are exposed. But for all the other admin sections, 
🟡 7. **Configure Community:** It could be helpful to indicate the widget type is associated with each add on... right now it's unclear. 
🟡 8. **Telegram Tab:** Since we're now allowing people to add multiple telegram handles, we should disable the ability to add comma-separated Telegrams. We can solve this later in the form validation ticket.
🟠 9. **Wiki Tab:** Since we have the preview, can we also show the Title & Description? This way the user will see exactly how the page will render.
🔴 10. **Wiki Tab:** Can we make the content of the page stretch 100% instead of 80%? 
🟠 11. **Wiki Tab:** Can we allow the admin to customize if to indent the title + description left, middle, or right? If not, I think we should align it left by default for now to match the rendering of the description. Also, it's funny that the page title is smaller than H1 ...  
🔴 12. **Wiki Tab:** If I leave title/description empty and click save and confirm transaction, I then see "No Wiki Configured" message. We should make title/description required to avoid this.
🟠 13. **Configure Community:** Can we add a character limit to title & description, similar to the other admin sections?

elliotBraem:
@ori-near The way you've organized this is excellent. Thank you, here is my response:

1. I have not been able to reproduce this, it happens immediately each time for me. When a community configuration page loads, it grabs all community data at once, including all of the addons -- and we don't have an indexer that tracks our changes, it's direct to and from the contract: so there's not really time that's needed to "update"... but it seems from your screenshots that you've been able to reproduce consistently... Perhaps this could be saved for a separate ticket and include the wallet that is being used, that may be a factor. 

2. ✅ 
3. ✅
4. ✅
5. The reason for this is because I'm disabling the up arrow on the top element and disabling the down arrow on the bottom element. I've opted for removing the option instead, how is this?
6. ✅
7. ✅
8. Cool, separate issue sounds great.
9. ✅
10. ✅
11. ✅
12. ✅
13. ✅

Check out the updated version in the [preview](https://near.org/devhub.efiz.near/widget/app?page=community&handle=devhub-test)

ori-near:
Hi @elliotBraem – thank you for addressing all my feedback so quickly. Restarting numbering with a new list of issues:

## Feedback Round 2

1. **Configure Community**: The Add-Ons section still displays the input field when not in edit mode. Ideally before a user clicks edit, that section will show up in a similar format to the other ones (Community Information, About, Access Control, etc.) Something like this:
2. **Configure Community**: Did we end up adding a limit for how many widgets one could add?
Screenshot 2023-10-27 at 11 23 20 AM
Once the user clicks edit, they'll see the editable fields.
3. **Configure Community**: Instead of having the question mark button under actions, could we add a new column before Tab Name that says Tab Type. And then just list Wiki/Telegram etc? Then the second column would say Tab Name.
4. **Wiki**: It seems like we have two paddings of 20px, resulting in a total padding of 40px. Can we remove one of those? So that we only have 20px padding and leave the content higher up on the page?
5. **Wiki:** The left/center/right buttons also shows up in the preview. Can we hide them in the preview?

Issue #268, opened 2023-09-20
Introduce new Blog plugin


ori-near:
**Problem**
Communities lack a dedicated plugin for displaying content such as community blog posts or announcements.

**User Story**
As an admin, I need a dedicated blog plugin to display labeled content in a feed-like manner for my community.

**Acceptance Criteria**
- Introduce a new Blog plugin that aligns with the new framework (#253). So an admin can enable, configure, or remove the plugin in the community settings page.
- If a community admin enables the Blog plugin, the Blog tab shows up on the community page navigation. And the page itself is populated with the Blog content set in the Blog plugin.
- If an admin enables the Blog plugin, they can click configure to customize the the tab name as well as the Blog specific fields, including:
  - Desired tags (comma-separated) for inclusion and exclusion.
  - Display type: Feed or grid
- Community admin should have a way to preview Blog from the configure settings. 
- If an admin enables the Blog plugin, the Blog content field is required

jaswinder6991:
The current story addresses the need for a dedicated blog plugin to display blogs for community pages. I have created a story for a feature for the **creation** of blogs that the plugin would eventually display. Here is the story: https://github.com/near/neardevhub-widgets/issues/269

jaswinder6991:
We have decided to not do #269 for now as it introduces some complexity which might not be important for the mvp. Nonetheless I want to highlight that along with this current ticket which specifies display of a blog feed and its configuration on the community page on the basis of a blog tag (or something relevent), we also need to take care of the blog creation process, meaning how admins can create blogs. 
One way to do this is to create simple posts on the activity feed with the blog tag and then render the blog feed with the blog tag using the blog plugin from this ticket. 
Second way is to give a create blog button on the plugin itself which uses the underlying post_type only but the creator form is different than it is right now for usual posts. 
@elliotBraem @Tguntenaar can you share your views if the second way is a big overhead and takes more than a day?

PR #345, opened 2023-10-25
Blogs

elliotBraem:
_Resolves #268  and #269_

**Acceptance Criteria**

## Blog plugin
- [x] Introduce a new Blog plugin that aligns with the new framework. So an admin can enable, configure, or remove the plugin in the community settings page.
- [x] f a community admin enables the Blog plugin, the Blog tab shows up on the community page navigation. And the page itself is populated with the Blog content set in the Blog plugin.
- [x] If an admin enables the Blog plugin, they can click configure to customize the the tab name as well as the Blog specific fields, including desired tags (comma-separated) for inclusion and exclusion and a display type: Feed or grid.
- [x] Community admin should have a way to preview Blog from the configure settings.
- [ ] If an admin enables the Blog plugin, the Blog content field is required
--A blog plugin can be created, and then needs to be configured directly to add content--

## Blogging

- [x] Introduce a dedicated "Create Blog" component accessible to community admins.
- [x] Header Image: Allow community admins to upload a header image for the blog. Provide a default image option.
- [x] Title: Provide a field to enter the title of the blog.
- [x] Author: Include an author field where the admin can specify the author of the blog.
- [ ] Date Posted: Automatically capture and display the date when the blog is created or posted.
- [x] Community Association: Ensure that each blog is associated with a specific community, allowing admins to link the blog to their community.
- [ ] Editing and Deletion: Admins should have the capability to edit or delete the blogs they create.
- [x] Integration with Community Blog Plugin
- [x] The blogs created should not be displayed as regular posts on the activity feed.
- [x] Preview cards for the Blogs should be different than usual devhub posts.
- [ ] Preview card completed design
- [ ] Blog page completed design


elliotBraem:
### How should blogs be published?

There are three options:

1. **Label**: We use a "comment" post_type and attach a "blog" label to the post. We exclude this label from the Activity feed and make it exclusive to the blog plugin feed. This involves no change to the contract. We are currently using Near Query API to index our posts and so we would only need to ensure we can filter by multiple labels. It would be a quick and easy solution, but would not be a final solution.
2. **Post Type**: We create a new post_type dedicated to blog posts. What data should be in this new post type? We could save data to the contract as a stringified JSON for quick iteration. We would need to ensure we can use the query API to filter by post_type and this will likely require a change to our indexer. Will blogs always be on the devhub contract? 
3. **Social DB**


@ailisp could you please check out the above two options and confirm if any changes on query api would be needed for?

ailisp:
> @ailisp could you please check out the above two options and confirm if any changes on query api would be needed?

We just discussed in the meeting, either option would require a bit change in adding a bit query in frontend, but doesn't affect query api indexer code.

ailisp (approved):
Code overall looks great. Will further test on staging

PR #383, opened 2023-11-01
Fix feed UI

itexpert120:


src/devhub/page/community/index.jsx
```diff
@@ -300,6 +301,19 @@ return (
             </Link>
           </small>
         </div>
+        <Widget
```
elliotBraem, line 304: To fix the test, can you do:

{context.accountId && <Widget .... }

We only show post button if user is logged in

Issue #129, opened 2023-05-16
UI for access control to manage permissions


ori-near:
**Problem**

Currently the permission system is not editable in the UI. You can only see who can do with which labels on UI but as an moderator you cannot change the settings.

**User Story**
As an moderator, I want to edit teams, team members and their permissions in DevHub.

**Acceptance Criteria**
- When non moderator user logs in, they cannot see or access the edit permissions page
- When moderator user logs in, there is a menu navigation item that only they can see to enter the Teams page. Non moderatos cannot see this.
- When moderator user logs in, there is a button in the Teams page, to enter edit mode
  - Moderators can add and remove restricted labels
  - Moderators can create and delete teams
  - Moderators can add member and remove member from a team
  - Moderators can specify which team can do with which restricted label

ori-near:
@ailisp Please fill out this ticket

Tguntenaar:
I would like to add this

Tguntenaar:
@ori-near I have a few questions about this card. 
The smart contract allows for nesting of members and teams. 
It also allows for members have their own labels.

So my first question is does the UI need to reflect this flexibility or be more restrictive?

My guess is more restrictive could you verify this?

In that case we will only allow 2 layers: teams with members. 
And restricted labels will be team specific and not member specific

Looking at the [current team page](https://near.org/devgovgigs.near/widget/gigs-board.pages.Teams). Teams seem to have 1 label and that label always allows for editing, assign and unassign. Is that also always the case or should this be flexible in a way that a label can be or editing only. Or assign and unassign only?

If you prefer to have a call let me know.

frol:
Well, I would like to start with that [the contract code](https://github.com/near/neardevhub-contract/blob/8dd8e5d1ba53ed30b8ef2ad1c99da2ed8e8d76a0/src/access_control/mod.rs#L49) only allows to edit moderators if the transaction is signed by the contract account id itself (`devgovgigs.near` on mainnet), so even moderators cannot invite or edit permissions at the moment.

> So my first question is does the UI need to reflect this flexibility or be more restrictive?

I vote for having _very simple UI_ for this functionality, potentially even a separate page not linked from the DevHub UI for now. As it is only useful for a couple of people on the team. The most common action is: add a new member to an existing group.

ori-near:
Hey @Tguntenaar is this still blocked? I believe we answered your questions.

ori-near:
[...]

Issue #315, opened 2023-10-18
[Admin] Give DevHub moderators the ability to manage label permissions on the admin page


ori-near:
**Problem**
As part of https://github.com/near/neardevhub-widgets/issues/275, we introduced a dedicated admin page to allow DevHub moderators to set featured communities through the UI. We want to centralize any site-level admin settings on that page. Currently, there is a separate page/UI for managing labels (https://github.com/near/neardevhub-widgets/issues/129) – but it is not intuitive to use or easy to access. We need to incorporate the label/team permissions functionality into this new admin/settings page. So that there's just one admin page with all the relevant settings.

**User Story**
As a DevHub Admin, I want to find the label permission UI + the feature communities in one UI on the homepage.

**Acceptance Criteria**
- An "Admin" link is visible in the header navigation, visible and accessible only to DevHub admins (not community admins!)
- When a DevHub admin clicks the Admin button, they see a settings page with the existing section for configuring the featured communities + a new section for configuring the label permissions
- Make the UI more intuitive for label permissions so that it is clear for a moderator how they can perform the common tasks, such as:
  - Seeing, adding, editing, and removing restricted labels
  - Seeing, adding, editing, and removing teams
  - Seeing, adding, editing, and removing members from teams
  - Configuring what teams can do with restricted labels 

  Note: Please consult with @Tguntenaar before working on this ticket to get more context on some of the current usability issues with the existing UI. And update this ticket accordingly.

Tguntenaar:
@ori-near I've put it in review

1. I've made a distinct difference between the admin team and other teams. 
2. I've removed the nonsense from the image below and simply replaced by the description of the admin team.

Let me know what adjustments need to be made to finish the issue.

![Image](https://github.com/near/neardevhub-widgets/assets/28901891/cf5c6d11-5e9f-459c-aeb0-1c2252e111ef)


[PREVIEW](https://near.org/geforcy.near/widget/app?page=admin)

Additional info in the [PR](https://github.com/near/neardevhub-widgets/pull/388)

Tguntenaar:
@ori-near this preview also contains issue #275 on top of the admin page. Shouldn't be an issue while testing and reviewing.

ori-near:
Hi @Tguntenaar – thanks for posting this.  The preview overall looks really good. I think the user permissions are much more clearly organized. 

# Feedback Round 1

1. **Moderators section:** 
a. Change title to "Moderators"
[...]

Issue #275, opened 2023-09-21
[Communities] Give DevHub moderators the ability to set Featured Communities on homepage (FRONT END)


ori-near:
**Problem**
As part of https://github.com/near/neardevhub-contract/issues/36, we made it possible to set the featured communities on the homepage on the contract side. However, admins do not have the ability to set the featured communities through the UI. Currently, they have to rely on developers to modify the smart contract.

**User Story**
As a DevHub admin, I need an easy way to update the features communities on the homepage so that I can showcase various communities without relying on developers to change the code.

**Acceptance Criteria**
- An "Admin" link is visible in the header navigation, visible and accessible only to DevHub admins (not community admins!)
- When a DevHub admin clicks the Admin button, they see a settings page with section for configuring the featured communities
- Within the settings section, DevHub admins can **edit** communities to display on the homepage. The interface for this can be super simple ... just need to solve the goal.
- The UI show a maximum of 5 featured communities, to support the current display limit
- Once the admin clicks save, the changes are instantly reflected on the homepage
- The admin UI should prevent the user from accidentally showing fewer than 4 communities.

Note: It would be great if we can incorporate the label/team permissions functionality (https://github.com/near/neardevhub-widgets/issues/129) into this new admin/settings page. So that there's just one admin page with all the relevant settings.

carina-akaia:
Hi @ori-near, would a community handle input field be enough for adding communities to the list of featured?

ori-near:
Hi @carina-akaia – yes, that's fine.

ori-near:
@carina-akaia Here we go:
https://github.com/near/neardevhub-widgets/issues/315

carina-akaia:
> @carina-akaia Here we go:
> 
> https://github.com/near/neardevhub-widgets/issues/315

Would you like me to take it later?

carina-akaia:
## Implementation demo

- [Mainnet](https://near.org/carina.akaia.near/widget/gigs-board.pages.Feed?nearDevGovGigsContractAccountId=devgovgigs.near)

carina-akaia:
@ori-near please perform QA using the demo ( notice that it's connected to the production contract! ) and give me your feedback

ori-near:
Hi @carina-akaia – I took a look at the main net link you sent but don't see the admin page in the navigation menu. Can you please check? Thank you.
[...]

PR #388, opened 2023-11-01
feat: configurator team, label, permission&member

Tguntenaar:
_Resolves #315_

[PREVIEW](https://near.org/geforcy.near/widget/app?page=admin)
comment down below if I need to add you to the moderators

Acceptance Criteria:

- [x] An "Admin" link is visible in the header navigation, visible and accessible only to DevHub admins (not community admins!)

- [x] When a DevHub admin clicks the Admin button, they see a settings page with the existing section for configuring the featured communities + a new section for configuring the label permissions

- [x] Make the UI more intuitive for label permissions so that it is clear for a moderator how they can perform the common tasks

In my last call with ori about issue #129 we discussed that I should rethink the formally known Teams page into an intuitive UI. I have made two major improvents and some minor:

1. We decided to simplify the task by only allowing 1 label per team. This made possible that the user can create a label and team at the same time instead of seperate and having to link them together afterwards.

2. The next improvement I made is removing most of  the buttons for every action. Before we had a create, an edit and delete button/toggle for every members, labels, teams, permissions. Now this all abstracted away by simply creating, editing and deleting a team in one configurator. 

3. We decided to make a clear distinction between the moderators group and other groups.

Additionally:
The moderators description (in the devhub.near contract) should be changed to:
The moderator group has permissions to edit any posts, and apply all labels including restricted ones.


src/devhub/entity/team/TeamInfo.jsx
```diff
@@ -0,0 +1,245 @@
...
+  useLabels: uslbls,
+  members: mmbrs,
+}) {
+  let txn = [];
+  let numberOfChanges = 0;
+
+  if (teamName !== tmnm) {
+    numberOfChanges++;
+    if (allTeamNames.includes(`team:${tmnm}`)) {
+      // Error team already exists
```
Tguntenaar, line 70: Still have to do this error

src/devhub/page/admin.jsx
```diff
@@ -0,0 +1,268 @@
...
+  description,
+  label,
+  editPost,
+  useLabels,
+  members,
+}) {
+  let txn = [];
+
+  if (rootMembers.includes(`team:${teamName}`)) {
+    // Error team already exists
```
Tguntenaar, line 100: Still have to do this as well

src/devhub/page/admin.jsx
```diff
@@ -0,0 +1,268 @@
...
+    },
+  ]);
+}
+
+const [createTeam, setCreateTeam] = useState(false);
+
+return (
+  <Container>
+    <div className="d-flex flex-column gap-4 p-4">
+      {/* {featuredCommunityList ? (
```
Tguntenaar, line 163: This is issue #275 which I will  follow up after this PR.

Tguntenaar:
Hi @ailisp! Thank you for reviewing. As I mentioned in the issue [here](https://github.com/near/neardevhub-widgets/issues/315#issuecomment-1823557854). I still had to update the preview with the latest changes which I'm planning to do today.

Tguntenaar:
Hi @ailisp! It is ready for review, I have updated the tests and the [preview](https://near.org/geforcy.near/widget/app?page=admin).

ailisp:
@Tguntenaar I tested, UI and the functionalities to edit label permissions look great! There're some bugs:
- After add/remove a featured community, redirect back from wallet, it's succeed but not reflected on the UI. The changes only visible after refresh.
- When I try adding `boatnear.near` as moderator, I got this error:
```
{"index":1,"kind":{"ExecutionError":"Smart contract panicked: panicked at 'Member does not exist', src/access_control/members.rs:201:52"}}
```

Tguntenaar:
@ailisp Hmm that's strange I did push a fix for that exact error. I will look into it and fix it.

Tguntenaar:
@ailisp Hmmm interesting, I did address this specific error in my code, but it seems it can still occur. I'll investigate further and ensure to resolve it by today

Tguntenaar:
@ailisp I did some testing and can reproduce your error. It seems that after a transaction is made, the getAllCommunitiesMetadata function is called 4 times total. First time it returns null, than an array with the length of 4 (old data) and than the updated array with a length of 5 (after the transaction data). However somehow the useState hook trips and the visible data on the page shows the array of 4. I've not been able to resolve this issue, any ideas?

Tguntenaar:
@ailisp Hmm, I might have overwritten the preview with another version. I deployed again and it worked for me. Could you try it once more?

Screenshot 2023-11-30 at 19 15 27

ailisp:
@Tguntenaar I have trouble to test it as currently wallet seems having trouble to connect to RPC nodes (certainly not related to your PR)
image

I'll try again later

ailisp:
Today wallet fails due to network congestion is resolved. This problem also doesn't exist now:
> @ailisp I did some testing and can reproduce your error. It seems that after a transaction is made, the getAllCommunitiesMetadata function is called 4 times total. First time it returns null, than an array with the length of 4 (old data) and than the updated array with a length of 5 (after the transaction data). However somehow the useState hook trips and the visible data on the page shows the array of 4. I've not been able to resolve this issue, any ideas?

Maybe it's also due to network congestion or there was fix on the gateway side

Issue #387, opened 2023-11-01
Background color does not match the design


frol:
**Describe the bug**

https://github.com/near/neardevhub-widgets/blob/37f3fd9f4add94584c3a20ef962194d22aef950d/src/devhub/components/templates/AppLayout.jsx#L8

I believe it has to be `#f4f4f4`, and it looks better with this color than the current one.

**To Reproduce**
1. Go to https://neardevhub.org
2. Notice that the background does not match the design

itexpert120:
Resolved in https://github.com/near/neardevhub-widgets/pull/393

PR #393, opened 2023-11-01
fix background color to match figma design

itexpert120:


itexpert120:
Fix for: https://github.com/near/neardevhub-widgets/issues/387

PR #402, opened 2023-11-01
Feature/349 sunset gigs boards

Tguntenaar:
_Resolves #349_

 > [!IMPORTANT] 
> Should only be merged if we are fully migrated


frol:
I completely removed all the components in #436 and manually pushed small proxy pages, so this is not needed.

PR #435, opened 2023-11-03
fix: reverts change to new Configurator

elliotBraem:
Solves the issue of community configurations breaking

Screenshot 2023-11-03 at 6 01 20 PM


frol (approved):

frol:
I checked on production, it works fine now. Thanks!

PR #443, opened 2023-11-04
Fix: fixed external links in mobile navbar

jaswinder6991:


src/devhub/components/organism/Navbar.jsx
```diff
@@ -243,15 +243,27 @@ return (
...
-                >
-                  {it.title}
-                </MobileLink>
-              ))
+              link.links.map((it, idx) =>
+                it.href.startsWith("http://") ||
+                it.href.startsWith("https://") ? (
+                  <MobileLink
+                    key={`nested-link-${idx}`}
+                    className={link.href === props.page && "active"}
```
frol, line 251: External links cannot be "active", so this condition is useless, but I will merge it as a hotfix and leave the whole block for refactoring after NEARCON.

frol (approved):

PR #452, opened 2023-11-09
Patch/448 post editing

### LINKED ISSUE #448: Edit and Reply for a post does not work on the post page.
**Describe the bug**
When you try to edit or reply to a post on the feed, it works fine and as expected but in the full window view of the post, the editing and reply on the post does not work.

**To Reproduce**
1. Go to a sample post [page](https://near.social/devhub.near/widget/app?page=post&id=2058 ).
2. Click on 'Reply'
3. Scroll down and try to submit your reply
4. Clicking on Submit button would result in no action. If you check console log, you will see the error like the below attached screenshot.

You see the same bug for editing the post too. The submit button won't do anything.

**Expected behavior**
- A reply should be posted successfully to the post.
- Author should be able to edit the post.

**Screenshots**
bug

**Additional Information**
A workaround for editing the post and replying to it is directly via the Activity feed. On the Activity Feed page, you can reply to the post and also edit it as well. This bug only occurs when the post is open on a full page.

carina-akaia:
Fixes #448 

# Demo environment
<https://near.org/devhub.jass.near/widget/app?page=home>


package.json
```diff
@@ -12,7 +12,7 @@
     "fmt:check": "prettier --check '**/*.{js,jsx,ts,tsx,json}'",
     "build": "npm run fmt && node ./module/include.js",
     "dev:testnet": "./scripts/dev.sh",
-    "dev:mainnet": "./scripts/dev.sh -a devgovgigs.near -c devgovgigs.near -n mainnet",
+    "dev:mainnet": "./scripts/dev.sh -a devhub.near -c devhub.near -n mainnet",
```
frol, line 15: The contract is still under devgovgigs.near account:

```suggestion
    "dev:mainnet": "./scripts/dev.sh -a devhub.near -c devgovgigs.near -n mainnet",
```

carina-akaia: @frol thank you!

frol (approved):
Seems to work fine

Issue #463, opened 2023-11-13
Add git hook for linting


elliotBraem:
**Is your feature request related to a problem? Please describe.**

More often then not, when working on a PR and pushing to the branch, the "Prettier" Check will fail because of not formatting before pushing. Then, the dev needs to create a separate commit just to fix formatting issues.

**Describe the solution you'd like**

This issue can be resolved with a git hook that runs the lint command, so that code is always formatted before being pushed to branch. With this in place, the check will always pass and we don't have to worry about the back and forth.

A possible (popular) solution is to use something like [Husky](https://typicode.github.io/husky/).
Add a pre-commit hook for running `npm run fmt`

**Describe alternatives you've considered**

I'm sure there may be other tools, although I've always used Husky, I'm also familiar with the complaints associated with its recent versions. Maybe an alternative like [lefthook](https://github.com/evilmartians/lefthook) or [simple-git-hooks](https://github.com/toplenboren/simple-git-hooks).

PR #464, opened 2023-11-13
feat: run fmt before push

itexpert120:
resolves https://github.com/near/neardevhub-widgets/issues/463


elliotBraem (approved):
I think this is a nice addition, thank you

Issue #456, opened 2023-11-10
Move to typescript


Tguntenaar:
**Is your feature request related to a problem? Please describe.**

While I was working on an issue trying to refactor some code I found myself missing typescript. I don't think I need to explain the clear benefits of typescript to our team. I think adding some it will improve code quality, maintainability and overall productivity.

**Describe the solution you'd like**

As mentioned by frol in [this post](https://t.me/devhubplatform/2/3054) on telegram - we could use his [typescript starter](https://github.com/frol/bos-component-ts-starter) 

**Additional context**

I'm trying to edit the package.json + scripts so that `npm run build` replace and transpile everything. Then we can add a watcher to re-run on file changes

Tguntenaar:
Ready for review! Check out the [PR](https://github.com/near/neardevhub-widgets/pull/468) here

Tguntenaar:
Addressed all comments of the first review.

Tguntenaar:
Update: After using the `watchman` nodejs package instead of the homebrew package. I needed to clean up the developer experience of running the project for development.

I was trying to instead of having all those individual scripts calling each other via npm run have it run as one application.

Here is a drawing of how it is supposed to work.

![Image](https://github.com/near/neardevhub-widgets/assets/28901891/967e3d75-9a5b-44e1-9afc-78da751328ca)

But ran into some trouble with ts-node so will try to make that work with plain javascript first.

Tguntenaar:
The development process has become a bit messy with some unneccesary complications in it. I've been looking into [bos-workspace](https://github.com/NEARBuilders/bos-workspace) with @elliotBraem yesterday. I will add Typescript to that solution so we can use it at DevHub aftwards. Making the development process a lot more straight forward. (develop, build, deploy, upload all in a single command)

Tguntenaar:
Update:  I've decided that moving to bos-workspace is out the scope of this ticket in which case it is ready to be merged once approved. See my latest comment in the PR for more context.

Tguntenaar:
update: some comments on the PR yesterday that I didn't have enough time for today, so will finish that tonight or tomorrow.

Tguntenaar:
Blocked, running `npm run watch` from the dev.sh script rebuilds too often if files change.

Tguntenaar:
[...]

PR #468, opened 2023-11-14
initial typescript support 

Tguntenaar:
_Resolves [#456](https://github.com/near/neardevhub-widgets/issues/456)_

I've used [this TS starter](https://github.com/frol/bos-component-ts-starter/) from @frol to support typescript in this repo. The process is as follows:
- copy all jsx files to build/src
- transpile the tsx file to jsx files to build/src
- replace Account Id Contract id and network with replace.js
- replace exports with build.js

This way we don't have to move everything to typescript now but we can do it file by file


Tguntenaar:
@frol @elliotBraem any feedback on this?

replacements.mainnet.json
```diff
@@ -1,7 +1,7 @@
 {
   "REPL_DEVHUB": "devhub.near",
   "REPL_DEVHUB_LEGACY": "devgovgigs.near",
-  "REPL_DEVHUB_CONTRACT": "devgovgigs.near",
+  "REPL_DEVHUB_CONTRACT": "thomasguntenaar.near",
```
frol, line 4: Just a friendly reminder: Don't forget to revert it before merging this PR.

.gitignore
```diff
@@ -6,4 +6,5 @@ node_modules
 replacements.*.json.tmp
 
 /build
+dist
```
frol, line 9: I would put everything into /build folder by creating more nesting there. having `build` and `dist` folders at the root of the project is confusing.

global.d.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+declare const Widget: (params: {
+  src: string;
+  props: object;
+}) => React.ReactNode;
+
+declare interface BaseProps {
+  page: any;
+}
```
frol, line 14: This looks like platform BaseProps, but these are DevHub BaseProps. Consider renaming it, adding a comment, or use explicit import instead of relying implicitly on this type.

frol (approved):
Overall, looks great to me, though i feel we also need to add some watch-io script command to package.json, so during local development we don't need to manually rebuild after each edit.

scripts/build.js
```diff
@@ -0,0 +1,70 @@
...
+
+  const packageJson = JSON.parse(
+    fs.readFileSync(new URL("../package.json", import.meta.url))
+  );
+
+  // await replaceInFiles({
+  //   files: [`${transpiledPathPrefix}/**/*.jsx`],
+  //   from: /^/m,
+  //   to: `/*\nLicense: ${packageJson.license}\nAuthor: ${packageJson.author}\nHomepage: ${packageJson.homepage}\n*/\n`,
+  // });
```
elliotBraem, line 55: I like this concept of stamping attributions and I think we should keep it; although I wonder if we could save it to metadata instead?

Maybe a feature saved for a [later iteration](https://github.com/NEARBuilders/bos-workspace/issues/2)

scripts/dev.sh
```diff
@@ -85,4 +85,4 @@ else
 fi
 
 # Run bos-loader with updated replacements
-~/.cargo/bin/bos-loader "$ACCOUNT_ID" --path src -r "$REPLACEMENTS_JSON.tmp"
+~/.cargo/bin/bos-loader "$ACCOUNT_ID" --path dist -r "$REPLACEMENTS_JSON.tmp"
```
elliotBraem, line 88: This should still be src

src/devhub/components/atom/Tag.tsx
```diff
@@ -0,0 +1,18 @@
...
+  const Span = styled.span`
+    color: ${black ? "#818181" : "#00ec97"};
+    font-size: 16px;
+    font-style: normal;
+    font-weight: 700;
+    line-height: 20px; /* 125% */
+  `;
+
+  return <Span>{tag}</Span>;
+};
```
elliotBraem, line 18: I think you should refactor this to:

```javascript

export default function Tag(props: TagProps) {
  const { black, tag } = props;

  const Span = styled.span`
    color: ${black ? "#818181" : "#00ec97"};
    font-size: 16px;
    font-style: normal;
    font-weight: 700;
    line-height: 20px; /* 125% */
  `;

  return <Span>{tag}</Span>;
};

```

And the above should work

.gitignore
Tguntenaar: Yea I will move it to build/src as you mentioned on telegram

scripts/build.js
```diff
@@ -0,0 +1,70 @@
...
+    files: [`${transpiledPathPrefix}/**/*.jsx`],
+    from: /import .* from "@\/includes\/([^"]*)";/gms,
+    to: (_match, importPath) => {
+      const importedFileContent = fs.readFileSync(
+        `${transpiledPathPrefix}/../includes/${importPath}.jsx`,
+        "utf8"
+      );
+      return `/* INCLUDE: "includes/${importPath}.jsx" */\n${importedFileContent}/* END_INCLUDE: "includes/${importPath}.jsx" */`;
+    },
+  });
```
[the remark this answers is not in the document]
frol: @elliotBraem Yes, it should be replaced with `VM.require`. I just did not have time to finish it.

scripts/build.js
frol: Good point about metadata. Let's do that.

scripts/dev.sh
frol, line 88: It would be great if we migrate all these replacements to `build.js` and potentially even avoid using bos-loader replacements in favor of some more natural syntax.

Tguntenaar: I agree, that would make it more straight forward

Tguntenaar:
> Overall, looks great to me, though i feel we also need to add some watch-io script command to package.json, so during local development we don't need to manually rebuild after each edit.

Thank you for the review. Which package are you referring to? I would use [watchman](https://facebook.github.io/watchman/), let me know if you have a better solution. I couldn't find watch-io.

scripts/build.js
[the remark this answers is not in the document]
Tguntenaar: I removed all `/* INCLUDE:` parts and moved to `VM.require`

scripts/build.js
Tguntenaar: Will keeps the comments for later use

scripts/dev.sh
Tguntenaar, line 1: The replacement logic is now located in `scripts/replace.js` since that works from both mac and on linux.

src/devhub/entity/community/Tile.jsx
Tguntenaar, line 1: Removed some unused variables and all of the leftover `INCLUDE:` comments to simplify `script/build.js`. This component is LEGACY and also should be replaced by `src/components/molecule/Tile` but outside the scope of this ticket.

elliotBraem (commented):
Running it locally:

Remembered to run ```npm install```
I got to near.org with flags set to ```http://127.0.0.1:3030```
And I'm getting Bos loader errors, along with a lot of this in the console:

```
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/main.rs:124:36
```


Investigating:

If I go to ```http://127.0.0.1:3030/```, nothing comes up. Looks like widgets aren't being served. 

I also don't see anything in my /build.


I tried deleting node_modules, /build, and /dist, reinstall and try again, still same issue.


If I run ```npm run build```, then /build/src/devhub/components/atom/Tag looks good! Although it still says "Main Component" and should say "Tag" instead (also there is a lot of whitespace):

```javascript
// white space
// white space
// white space
// white space
// white space
function MainComponent(props) {
  const { black, tag } = props;

  const Span = styled.span`
    color: ${black ? "#818181" : "#00ec97"};
    font-size: 16px;
    font-style: normal;
    font-weight: 700;
    line-height: 20px; /* 125% */
  `;

  return <Span>{tag}</Span>;
}

return MainComponent(props, context);
```

I also see that a new directory has been created: "/neardevhub-widgets"

scripts/build.js
```diff
@@ -0,0 +1,58 @@
...
+
+const transpiledPathPrefix = "./build/src";
+
+async function build() {
+  await replaceInFiles({
+    files: [`${transpiledPathPrefix}/**/*.jsx`],
+    from: /export\s+default\s+function[^(]*\((.*)/gms,
+    to: (_match, rest) =>
+      `function MainComponent(${rest}\nreturn MainComponent(props, context);`,
+  });
```
elliotBraem, line 12: MainComponent should be the name of the component, try something like:

```javascript
  await replaceInFiles({
    files: [`${transpiledPathPrefix}/**/*.jsx`],
    from: /export\s+default\s+function\s+(\w+)\((.*)/gms,
    to: (_match, funcName, rest) =>
      `function ${funcName}(${rest}\nreturn ${funcName}(props, context);`,
  });
  ```

**Comment by petersalomonsen** [comment]
So from my testing so far in Codespaces ( Linux ), you need to install watchman ( using brew ) and `cpio`. If possible it would be good to use javascript to obtain the same so that an `npm install` is sufficient. Otherwise we need clear instructions that informs about what needs to be installed, or a devcontainer.json for vscode / github codespaces.

**Comment by Tguntenaar** [comment]
Thanks Peter and Elliot for the review. I'll make the necessary adjustments to the code.
- I'll switch to the Node.js package from watchman to streamline the process within Codespaces.
- I'll transition away from using `cpio` and make use of `cp` instead, ensuring a more straightforward setup.

**Comment by Tguntenaar** [comment]
Update: on Mac the watchman package will throw a `Operation not permitted` error the first time if the project is stored inside the Documents folder. Which can be resolved by running `npm run clear:watch` which I added to the watch script.

More info [here](https://github.com/facebook/watchman/issues/977)

File: scripts/build.js (hunk 8 above)
Comment (Tguntenaar) [reply scripts/build.js:12]: That works works! No more white space and a proper function name
Screenshot 2023-11-20 at 11 09 49

**Comment by Tguntenaar** [comment]
@elliotBraem luckily all your problems are related and should be an easy fix. The `tokio-runtime-worker` paniced because the bos-loader is serving from the `build/src` folder. That isn't created for you but instead `/neardevhub-widgets` I think copying the files with `cpio` might have been the issue. I now just the `cp` command instead. So could you try again?

File: scripts/build.js [hunk 9]
```diff
@@ -0,0 +1,58 @@
...
+  await new Promise((resolve) => {
+    fs.rename(
+      transpiledPathPrefix,
+      `${transpiledPathPrefix}/../../${packageJson.name}`,
+      () => {
+        resolve();
+      }
+    );
+  });
+
```
elliotBraem, line 54: This is the reason for the neardevhub-widgets directory

scripts/build.js
```diff
@@ -0,0 +1,58 @@
...
+  // with empty content includes.
+  await new Promise((resolve) => {
+    fs.rename(
+      `${transpiledPathPrefix}/includes`,
+      `${transpiledPathPrefix}/../../includes`,
+      () => {
+        resolve();
+      }
+    );
+  });
```
elliotBraem, line 33: We can remove this

ailisp (approved):
This is really nice setup now! Looks very good to me

scripts/build.js
Tguntenaar: Thanks!

frol (changes requested):

scripts/dev.sh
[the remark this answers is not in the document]
Tguntenaar: Good point I will do that

itexpert120:
This is looking very good, is the intellisense working with TypeScript files?

scripts/dev.sh
```diff
@@ -54,35 +30,5 @@ while [[ $# -gt 0 ]]; do
...
-        jq --arg CONTRACT_ID "$CONTRACT_ID" ".[\"$CONTRACT_REPL\"] = \"$CONTRACT_ID\"" "$REPLACEMENTS_JSON.tmp" > temp.json && mv temp.json "$REPLACEMENTS_JSON.tmp"
-    fi
-else
-    echo "Error: $REPLACEMENTS_JSON file not found."
-    exit 1
-fi
-
 # Run bos-loader with updated replacements
-~/.cargo/bin/bos-loader "$ACCOUNT_ID" --path src -r "$REPLACEMENTS_JSON.tmp"
+~/.cargo/bin/bos-loader "$ACCOUNT_ID" --path build/src
```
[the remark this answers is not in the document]
Tguntenaar: I'm trying to make this work but it seems that the watcher will fire 8 times on every change when trying to execute it from the same script. @frol @elliotBraem  I just pushed what I've got, do you guys have any idea how to solve it elegantly?

Tguntenaar (commented):
This PR is ready for review again. Yesterday I had issues with the watch script that it would hot reload 8 times on changes to Tag.tsx. Today these issues seem to be gone at my end. Could you thoroughly test this?

CONTRIBUTING.md
```diff
@@ -28,12 +29,26 @@ npm install
 
 #### Run the dev script
 
-To initiate the development environment, you can use the following script. This script will first check if bos-loader is installed and, if not, install it. Then, it will serve the default development environment.
+To initiate the development environment, you can use the following script. This script will first check if bos-loader is installed and, if not, install it. Then, it will serve the default development environment and watch for changes inside `./src`.
 
-```sh
+```zsh
```
frol, line 34: nit: I believe `sh` is generic enough, so I don't see a justification for `zsh`. I don't want to have a mix of "sh" in the READMEs, so I suggest to keep them all `sh`.

frol (approved):
It looks good to me, but I have not tested it locally.

@Megha-Dev-19 Could you give it a try and see if it works as expected when you edit `src/devhub/components/atom/Tag.tsx`? 🙏

Megha-Dev-19:
hey @frol @Tguntenaar, I just tested it locally and the changes are reflected as expected.  👍

ailisp:
@Tguntenaar Can you resolve the conflict and we'll merge it!

elliotBraem:
Waiting to resolve conflicts then will merge

scripts/dev.sh
```diff
@@ -5,45 +5,21 @@ command_exists() {
...
-            ;;
-        -c|--contract)
-            CONTRACT_ID="$2"
-            REPLACE_CONTRACT=true
-            shift
-            shift
-            ;;
-        -n|--network)
-            NETWORK_ENV="$2"
+            ACCOUNT_ID="$2"
```
elliotBraem, line 22: @Tguntenaar 

Getting rid of the custom contract and network has made the following command stop working:

```cmd
npm run dev:testnet -- -a devhub-dev.testnet -c bodevhub.testnet -n dev
```

The account should be custom because it can either be devhubtest.testnet or devhub-dev.testnet (so package.json should not hardcode the account)

Contract needs to be custom, for example when using bo's contract for announcements

Network should be custom to follow contract and account

package.json
```diff
@@ -1,25 +1,35 @@
...
-    "build": "npm run fmt",
-    "dev:testnet": "./scripts/dev.sh",
-    "dev:mainnet": "./scripts/dev.sh -a devhub.near -c devgovgigs.near -n mainnet",
-    "build:preview": "node ./scripts/build-preview.mjs",
+    "build": "npm run fmt && npm run copy:jsx && npm run transpile && npm run replace && node ./scripts/build.js",
+    "copy:jsx": "mkdir -p build/src && cp -r ./src ./build",
+    "transpile": "sucrase ./src -d ./build/src --transforms typescript,jsx --jsx-runtime preserve --disable-es-transforms --out-extension jsx",
     "dev": "npm run dev:mainnet",
+    "dev:mainnet": "./scripts/dev.sh -a devhub.near",
+    "dev:testnet": "./scripts/dev.sh -a devhub.testnet",
```
elliotBraem, line 28: Just "dev:testnet": "./scripts/dev.sh"

scripts/replace.js
```diff
@@ -3,11 +3,10 @@ import path from "path";
 
 const args = process.argv.slice(2);
 
-let NETWORK_ENV = "mainnet";
+let NETWORK_ENV = "testnet";
```
elliotBraem, line 6: Is this right? I think this should maybe default to mainnet or be overriden from arguments

scripts/dev.sh
Tguntenaar: Hey Elliot! Thank you for pointing this out. We should discuss this. 

The replacement of those constants in now handled in replace.js instead.

The bos-loader only uses the the ACCOUNT_ID to deploy the widgets.

So when you want to change the account / contract / network this can be done by either editing the package.json or making use of the replacements.*.json files.

If we want `npm run dev` to take those arguments we will have to pass it on to the `npm run watch` script, which has to pass it to the `npm run build`, which passes it to the replace... I thought this would be less optimal.

Btw this is because the watch has to rerun the replace script for every build.

Issue #479, opened 2023-11-16
Hide tickets with the "devhub-test" label from the activity feed


ori-near:
# Problem
Currently, DevHub portal developers sometimes need to create test posts. However, these tend to clutter the activity feed:
![Image](https://github.com/near/neardevhub-widgets/assets/110252255/642e54f0-20ea-42bb-a9fb-acbd22156a29)

# User Story
As a DevHub portal developer, I want to be able to test posts without creating a distraction for users.

# Acceptance Criteria
- [ ] Hide posts with the `devhub-test` label from the activity feed

PR #400, opened 2023-11-01
Fix/376 blog posts should not appear in activity feeds

petersalomonsen:
based on https://github.com/near/neardevhub-widgets/pull/369

fixes #376

PR #486, opened 2023-11-16
Hide posts with "devhub-test" tag from feed

itexpert120:
Resolved https://github.com/near/neardevhub-widgets/issues/479

The posts with tag "devhub-test" are hidden from the feed by default, but they can be viewed when searching for "devhub-test" tag

### DIFF
--- a/src/devhub/entity/post/List.jsx
+++ b/src/devhub/entity/post/List.jsx
@@ -94,13 +94,22 @@ function getPostIds() {
     where = { parent_id: { _is_null: true }, ...where };
   }
 
-  // Don't show blog
+  // Don't show blog and devhub-test posts
   where = {
-    _not: {
-      labels: { _contains: "blog" },
-      parent_id: { _is_null: true },
-      post_type: { _eq: "Comment" },
-    },
+    _and: [
+      {
+        _not: {
+          labels: { _contains: "blog" },
+          parent_id: { _is_null: true },
+          post_type: { _eq: "Comment" },
+        },
+      },
+      {
+        _not: {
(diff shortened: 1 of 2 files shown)


src/devhub/entity/post/List.jsx
```diff
@@ -153,6 +153,8 @@ function defaultRenderItem(postId, additionalProps) {
           onDraftStateChange,
           ...additionalProps,
           referral: postId,
+          hideTestPost: props.tag !== "devhub-test",
```
elliotBraem, line 156: See the [PR here](https://github.com/near/neardevhub-widgets/pull/400/files#diff-0232ce282bd54ecead7451ede7726988c82b7fbdfc708465c44cde3788ae65d9) for how we hide Blog posts from the feed.

This can be done the same way

elliotBraem (changes requested):

src/devhub/entity/post/List.jsx
itexpert120: Ah thanks, but then how could they access it?

elliotBraem: I don't think we need them to be searchable from the main feed -- they can be ignored just like blogs are.

Posts with devhub-test will always be accessible from the [devhub-test community](https://near.social/devhub.near/widget/app?page=community&handle=devhub-test)

itexpert120: Got it, thanks for help. Let me fix this asap

elliotBraem (approved):

PR #488, opened 2023-11-18
Feat/137 avoid wallet confirmation

### LINKED ISSUE #137: Avoid a need to confirm every DevHub post/like through a Wallet
## Is your feature request related to a problem? Please describe.

There are two problems at the moment:

1. In order to post a new idea/solution/reply on DevHub or like a post on DevHub users are prompted to sign a transaction through a Wallet since we make a function call on the DevHub contract. This is invasive and breaks users' flow, and they are more hesitant with responding/linking.
2. DevHub [cross-posts top-level posts under its own Social Feed](https://near.social/#/mob.near/widget/ProfilePage?accountId=devgovgigs.near), and comments and likes left there are not displayed on DevHub.

## Describe the solution you'd like

Use SocialDB posts as the main source of truth to merge two worlds.

I suggest that posting on DevHub will create a regular SocialDB post visible on the feed to everyone who follows the author of the post, and the DevHub indexer would catch those posts and publish references to them to the DevHub contract.

This way we achieve the following:

* Support of fast-auth (free transactions for newly onboarded users)
* No interruptions while using DevHub as likes and posts are submitted to SocialDB using a function-call access key that the BOS gateway holds, and it does not need to redirect to Wallet to sign a transaction
* More engagement as users posting to DevHub will also post to their own feed and their followers will notice and participate in the discussions
* More likes (it would be a no-brainer)

## Describe alternatives you've considered

We may use SocialDB for the posts and avoid feed, but we won't benefit from users broadcasting knowledge themselves.

We may stick to using our own contract and sync missing likes and comments to DevHub, but that won't remove the friction from the user experience.

## Additional technical context

In order to post to Social Feed, users basically [override `<username>/post/main`](https://explorer.near.org/transactions/2vmat22ZzeVKddtEgcaDbXtoSe36rCLWGRKZgDiuUZkB), so when you look at the Post component props, https://near.social/#/mob.near/widget/MainPage.Post.Page?accountId=root.near&blockHeight=92478864, notice `blockHeight` in the URL - the component will basically make a function call to get the data from RPC at that particular historical moment, as SocialDB contract does not store all the posts in its storage, it overrides the data with each new post/comment.

Just for reference, [Here](https://explorer.near.org/transactions/ExJLjqLfUThTwTxYtabgVKNREXA97UCQyY4LozVqhfTi) is the comment response on the Feed, and [here](https://explorer.near.org/transactions/6Gd7FFz3dHPiNApoo7PAX1Aho6BsExJWtrLMF3ubdjsY) is like.

There is an `index` key that needs to be properly constructed, and I have not looked into it in more detail.

Here are the features that we need to take into account:

### Post creation

As I mentioned before, I suggest we call a `set` function on SocialDB with:

```json
{
  "<author-account-id>": {
    "post": {
      "main": "MESSAGE"
    },
    "index": { /* figure out how to construct it properly and potentially add DevHub-specific properties */ }
  }
}
```

It is worth noting that "MESSAGE" might be not only the message itself, but also a link to the DevHub thread discussion as Social Feed does not support nesting (see below).

### Reply through DevHub

I believe that each reply should be just another Social Feed post (not a comment), similar to how it worked on Twitter when users imitated threads and retweets ("RT") before they got integrated into the UI. DevHub replies should look like regular posts on Social Feed, but I would add a link to the parent post on DevHub and to the post in the thread on DevHub.

### Reply through Social Feed

Comments on Social Feed should be just picked up as comments to the relevant post on DevHub by the indexer.

### Nested responses

As mentioned above, I see DevHub posts to be represented by individual posts on Social Feed and in the message itself there should be a link to the DevHub thread. DevHub indexer would catch such posts and add them to the DevHub contract.

### Post editing

Social Feed does not support editing, so we will need to simulate that somehow. We might post edits as diffs into the comments (potentially try to hide in spoilers, but spoilers are not supported by Social VM). 

~~~
@dev-hub.near Apply the change to the original [message](#link-to-DevHub-post).

```
-#old-label #x
+#new-label #x

-Old text
+New text
```
~~~

### Permissions to edit posts

I think it will be up to the contract to decide whether accept the edits or not while diffs can be commented on Social Feed by anyone and we cannot restrict that.

## Task Breakdown

This ticket is a spike. It expects to take one engineer 1-2 week to complete each one of the milestone.

### Milestone 1. Proof of concept implementation on challenging features
First we want to ensure the most challenging part of SocialDB based approach is doable. Tackle and figure out technical difficulties in advance. If it's not feasible we can plan early and in the worst case terminate this plan. From discussions with @frol we think the most challenge parts to implement are:

- Simulate post editing with diff in comments
- Plan to migrate existing posts stored in contract to SoicalDB.

So let's experiment on these features first.

### Milestone 2. MVP of all features mentioned above

By the end of this milestone, user should be able to do everything with post as they can do in current DevHub board. It includes write, read, edit, reply, like, labeling, types and search posts, but doesn't include community features.

### Milestone 3. Migrate existing posts to new one

This milestone is about actually migrating the data from existing contract to SocialDB. Ideally Old posts should still comes from the user's account, but given we cannot represent user's account to batch create user's all history posts, a bot account that grant user's ability to edit would be acceptable. UX details is upon further discussion.

### Milestone 4. Integration and merge to production

By this milestone we consider complete this task. All post related functionalities in milestone 2 and all history posts should be deployed in production. User should be automatically interact with the SocialDB posts instead of DevHub posts. Note the scope of this spike doesn't include community in SocialDB, which will be our next step.

petersalomonsen:
work in progress to resolve #137 

this depends on https://github.com/NearSocial/VM/pull/156 where you can also find a preview environment


ailisp:
Hi @petersalomonsen ! What's the status of this ticket? Any blockers?

petersalomonsen:
Hi @ailisp 

There are no blockers, but all the work is currently going on in this PR https://github.com/NearSocial/VM/pull/156

You can find a preview environment in there as well.

Everything works, but there are two remaining things to conclude on:

-Avoid reloading the page after a completed transactions. My idea on how to sove this is to inject `transactionHashes` into the top level `props`
- When adding the function access key to the wallet, the message should be `an application is trying to get limited access to your account`. Currently you have to really pay attention and inspect the transaction to see that adding a function access key is part of what you are confirming in the wallet. I haven't looked into how to solve this yet.

My plan is to conclude of this by the end of this week ( by the end of the weekend ).

Please have a look on the current progress in the PR in the VM, and any comments on the current state are welcome. Would be great with your feedback too @nearmax

petersalomonsen:
https://github.com/NearSocial/VM/pull/156 could need some comments now. 

Also we need to fix this issue: #549

petersalomonsen:
This upgrades to the latest near-bos-webcomponent, that also supports the wallet selector, so would like to keep this PR for a little longer and see how we can test wallet confirmation related stuff.

petersalomonsen:
Will close this, as I believe it is not needed anymore. We have solved it in https://github.com/NearSocial/VM/pull/156, and we have tests for the "Don't ask again" feature here in the DevHub repository.

PR #508, opened 2023-11-23
Fix filter tag action

### LINKED ISSUE #429: Click on tag doesn't filter by tag any more
**Describe the bug**

In feed page click on any post's tag, e.g. nearcon, the url bar will rewrite with `tag=nearcon`: https://near.org/bo.near/widget/app?page=feed&tag=nearcon, however app doesn't realize this props change and not cause a re-render. The re-render only happens after manually refresh the page.

**To Reproduce**
1. Go to activity feed page
2. Click on any tag of any post
3. The url changes, but page doesn't filter by the tag

**Expected behavior**
The page should filter by the tag

**Additional context**

Elliot and I have come up with some potential fixes, but I was not able to fix it in a good way. 

Elliot's thoughts:
> Elliot, [11/2/23 10:44 PM]
I’ve been trying to think about this and I don’t really know besides setting the tag in state and passing that as a prop
Elliot, [11/2/23 10:44 PM]
So linking but also changing state on click
Elliot, [11/3/23 12:45 AM]
Thinking more about it, it may be that we need to update a key on the div that matches the tag maybe. Or the query parameters as a whole. I'll try something out; thanks for bringing this up

Bo's attempts:
tried out but the link to a tag is a `<Link>`, It seems not possible to set the target widget's state with it. A way may work is to escalate tag props all the way up to the app widget, and pass a updateTag function all the way down to the post-search/panel.jsx. Then use props.tag instead of state.tag.

Megha-Dev-19:
Resolves #429 
Preview [link](https://near.org/megha19.near/widget/app?page=feed)

### DIFF
--- a/src/devhub/entity/post/Post.jsx
+++ b/src/devhub/entity/post/Post.jsx
@@ -626,7 +626,7 @@ const tags = post.snapshot.labels ? (
       <div className="d-flex align-items-center my-3 me-3">
         <Link
           to={href({
-            widgetSrc: "${REPL_DEVHUB}/widget/app",
+            widgetSrc: "#/${REPL_DEVHUB}/widget/app",
             params: { page: "feed", tag: tag },
           })}
         >


ailisp (approved):
Looks great. Thank you!

PR #551, opened 2023-12-05
Removes deposit and reduces gas

elliotBraem:
_Resolves #549_

This removes the deposit from all function calls to DevHub contract in our codebase, and set the gas for all of them to `Big(10).pow(14)` ([MAX_GAS](https://github.com/near/near-api-js/blob/858b8ee35ffc7d6409b3d72b01f59f4e117db613/packages/cookbook/utils/calculate-gas.js#L10))


petersalomonsen:
Thanks @elliotBraem I will test this later today

petersalomonsen:
@elliotBraem I've tested it now, and it works!

Issue #557, opened 2023-12-07
Reply comment does not work (no `labels` field set)


frol:
**Describe the bug**

Here is the transaction that was generated by DevHub: https://legacy.explorer.near.org/transactions/6BtgfarYoikfGwLRnVWuZSqXnis7zLREtvWoXKtPRvRX

**To Reproduce**
1. Go to any post
2. Reply any text
3. Observe that the prepared transaction does not have `labels` field which is requires on the smart contract side.

**Expected behavior**

Either make the `labels` field optional on the contract side, or provide it from the UI.

PR #558, opened 2023-12-08
Fix empty labels error

### LINKED ISSUE #534: Label missing error in post creation
Steps to reproduce:
1) Create a post [here](https://near.org/devhub.near/widget/app?page=create)
2) Don't add a label and proceed
image

I think in UI we should add checks (about mandatory fields) before creating the txn

Megha-Dev-19:
Resolves #534 and #557 
Preview [link](https://near.org/megha19.near/widget/app?page=feed)


frol (approved):
I confirm that it works now (manually tested the preview link)

PR #567, opened 2023-12-14
feat: Community Announcements Page

Megha-Dev-19:
Preview [link](https://near.org/megha19.near/widget/app?page=community&handle=devhub-platform)

### DIFF
--- a/src/devhub/entity/community/Compose.jsx
+++ b/src/devhub/entity/community/Compose.jsx
@@ -143,7 +143,6 @@ function autoCompleteAccountId(id) {
   );
   State.update((lastKnownState) => ({
     ...lastKnownState,
-    handler: "autocompleteSelected",
     text: updatedDescription,
     showAccountAutocomplete: false,
   }));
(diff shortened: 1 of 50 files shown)


Megha-Dev-19:
Am I allowed to attach the figma design as well? for reference?

src/devhub/entity/community/Compose.jsx
```diff
@@ -0,0 +1,477 @@
+if (!context.accountId) {
+  return <></>;
+}
+
+const optimisticUpdateFn = props.optimisticUpdateFn;
+const clearOptimisticUpdateFn = props.clearOptimisticUpdateFn;
+
```
elliotBraem, line 7: Where did these two functions come from? It looks like optimisticUpdate is for immediately reflecting the newly created post?

Megha-Dev-19: yeah, we can use it for passing some custom function for commit and cancel as well

ailisp:
Looks good so far! I think it is blocked by https://github.com/NEAR-DevHub/neardevhub-contract/pull/79 to call smart contract to post from community account

PR #606, opened 2023-12-27
Enhance announcements UI

### LINKED ISSUE #605: Enhance announcements page
- [x] Improve the UI for posts list (like that of near.social page)
- [x] Update card style

Megha-Dev-19:
_Resolves #605_

### DIFF
--- a/src/devhub/entity/community/Announcements.jsx
+++ b/src/devhub/entity/community/Announcements.jsx
@@ -25,6 +25,16 @@ const SidebarContainer = styled.div`
   flex: 1;
 `;
 
+const Heading = styled.div`
+  font-size: 19px;
+  font-weight: 600;
+`;
+
+const SubHeading = styled.div`
+  font-size: 15px;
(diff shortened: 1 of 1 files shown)


src/devhub/entity/community/Announcements.jsx
```diff
@@ -105,15 +119,18 @@ return (
...
                 },
               ]}
-              Item={(p) => (
+              Item={(item) => (
                 <Widget
-                  loading={
-                    <div className="w-100" style={{ height: "200px" }} />
-                  }
-                  src="${REPL_MOB}/widget/MainPage.N.Post"
+                  src="near/widget/Posts.Post"
```
elliotBraem, line 124: Since we want this to work in both mainnet and testnet, could you please use replacements?

${REPL_NEAR}

These replacements are applied on build during the dev script, you can find the value it is replaced with in replacements.<env>.json. This allows the code to work across regions

PR #614, opened 2023-12-29
Fix Build error

### LINKED ISSUE #613: Build Error
Steps to reproduce: 
- `npm run build` throws error
image

Megha-Dev-19:
Resolves #613


src/devhub/entity/post/History.jsx
```diff
@@ -7,22 +7,7 @@ props.timestamp: number;
 props.referral: any;
 */
 
-function href(widgetName, linkProps) {
```
elliotBraem, line 10: With this replacement, did you double check that the function call is adjusted? This href function is slightly different from the VM.require href

Megha-Dev-19: I didn't since the code was from common.jsx, I expected it to be similar to href of VM
I will take a look

Megha-Dev-19: Fixed it, it works now

PR #615, opened 2023-12-29
Reuse PostEditor instead of Create component

### LINKED ISSUE #538: UX: Redirection after creating/editing post
Currently, when a use creates/edits a post, they just sit there on the same creation/edit page after the transaction is done (for me, using Meteor wallet). Having a consistent experience where the user is redirected to the created post or refreshing the feed after editing seems like a good idea

### LINKED ISSUE #547: Cancel button doesn't work for second level solution post replies
**Describe the bug**
As a follow up to the work done in https://github.com/near/neardevhub-widgets/issues/454, the cancel button doesn't work in comes cases for nested sponsorship post replies.

**To Reproduce**
1. Go to https://near.social/devhub.near/widget/app?page=post&id=2372
2. Click on reply --> Sponsorship
3. Click cancel
4. Editor doesn't cancel

**Expected behavior**
Editor should cancel anytime cancel is clicked.

 - OS: OS
 - Browser: Chrome
 - Version: 119.0.6045.199

### LINKED ISSUE #555: Refresh after approving txn using `Here wallet`
Steps to recreate:
- add a comment on any post
- sign txn using 'Here Wallet'
- on approval of txn, the page doesn't show me my comment, need to refresh to see it

### LINKED ISSUE #590: Create page should share the common "PostEditor" component
Although they implement nearly the same functionality, there is duplication and inconsistencies between the two of them (which has resulted in bugs, such as #583 and #576). We should condense these so the create page uses the reusable "PostEditor" component.

Test coverage has already been completed for the create page, these tests can be repurposed for the post editor.

### LINKED ISSUE #591: Reply or edit a post doesn't close up editor after successful submit
Now when you reply a post, and successfully submit it, you'll still see the editor open, with the content you just wrote. And the cancel button doesn't work:
image

## Acceptance criteria
- [ ] successful reply post should not show the editor
- [ ] successful edit post should not show the editor
- [ ] cancel button on editor should close the editor
- [ ] add tests two ensure above requirements and avoid regression

Megha-Dev-19:
Resolves #538
Resolves #555
Resolves #591
Resolves #590
Resolves #547
Preview [link](https://psalomobos.near.page/megha19.near/widget/app?page=feed)


Megha-Dev-19:
Redirection on Approval/Rejection of txn works only on @petersalomonsen preview, since it invalidates the cache which cause rerendering

Megha-Dev-19:
Since we can't redirect user without their explicit action in BOS, so I did this instead.
@ori-near please let me know if this is suitable UX

https://github.com/NEAR-DevHub/neardevhub-bos/assets/100185149/6a978957-ee6c-4aff-87d6-d7659176e47c

ori-near:
Hi @Megha-Dev-19 – It this UX going to apply to all users? Or just ones who use MNW? Can you also share the current state that you're trying to fix with this solution?

Megha-Dev-19:
Hey @ori-near, this UX will be applied to all users (once Peter's [PR](https://github.com/NearSocial/VM/pull/156) is merged), but until then only to those who use his MNV. 
So there are two things: one is when we edit/reply on a post, and when we create a new post. And for each there is two possibilities, user can use a wallet, which redirects them to new approval screen (a new link), or opens a popup modal.

Currently this is how it is handled: 

1) Post creation (when user is redirected for approval)

https://github.com/NEAR-DevHub/neardevhub-bos/assets/100185149/fd9edfa5-71f2-4a76-b20b-13193bf88773

2) Post creation (when popup window is opened for approval)

https://github.com/NEAR-DevHub/neardevhub-bos/assets/100185149/258aa588-4687-405e-a16a-d2513a32f743

3) Edit/reply post (when popup window is opened for approval)

https://github.com/NEAR-DevHub/neardevhub-bos/assets/100185149/73499431-db43-4381-adab-1cb83385cc5f


but with current changes, edit/reply post will be like this (irrespective of if user has allowed the function calls) :

https://github.com/NEAR-DevHub/neardevhub-bos/assets/100185149/45e80548-7600-4b34-b7a6-713f7526980b

ori-near:
Hi @Megha-Dev-19 – thank you so much for clarifying. As we discussed, I think this solution looks good. However, if there is a way to redirect the user to the actual post that would be even better.

CC: @petersalomonsen

petersalomonsen:
In order to redirect the user to the latest created post, we need the result of the create transaction, which should return the post_id. To get results from a transaction, we depend on this: https://github.com/NearSocial/VM/issues/105

Megha-Dev-19:
hey @Tguntenaar @elliotBraem, can you please review the PR.?

Megha-Dev-19:
hey @ori-near, as we discussed yesterday, I can't scroll to the top or the edited post, since we don't have access to `window` in VM, so it will remain as is for now.

PR #617, opened 2023-12-30
fix(test): flaky solution/sponsorship tests

petersalomonsen:
- parse and stringify transaction text content to avoid flakyness caused by whitespace diff
- use `locator` instead of `selectAndAssert`

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -36,6 +36,5 @@ jobs:
           npx playwright install
       - name: Run tests
         run: |
-          set -e
           npx playwright test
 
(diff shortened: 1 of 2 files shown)


Tguntenaar (commented):
Nice solution! Does the CI also fail when the tests fail now?

petersalomonsen:
> Nice solution! Does the CI also fail when the tests fail now?

The reason the CI did not fail was that theses tests were "flaky". There is a retry mechanism in Playwright, so if a test fails on the first attempt, but pass on the second, then the CI will pass: https://playwright.dev/docs/test-retries#retries

You can also see how we have configured this for the CI only with two retries: https://github.com/NEAR-DevHub/neardevhub-bos/blob/main/playwright.config.js#L29

PR #635, opened 2024-01-11
Fix no storage -> deposit issue

### LINKED ISSUE #629: Error confirming "grant_write_permission" transaction if the account has no storage deposit at social.near
**Describe the bug**

Given a new account that has not interacted with near.social before, and has no deposits for storage. When trying to interact with the DevHUB contract by for example adding a post. Then this involves two transactions, which besides the `add_post` function call on the DevHUB contract, there is also a transaction for `grant_write_permission` to `social.near`. The DevHUB frontend attaches 0.01 NEAR which results in the error message from the `social.near` contract that  `attached deposit is less than the minimum storage balance`. This message is from [internal_create_account](https://github.com/NearSocial/social-db/blob/d28c647252ce25a06c70c3b7f4930ccdcd217fd9/contract/src/account.rs#L135) in `social-db`.

**To Reproduce**
1. Use a new NEAR account with near.social, an account that has no previous interaction
2. Go to DevHUB activity feed and try to like any post
3. Notice the first transaction calling `grant_write_permission` with an attached deposit on 0.01 NEAR (which is not enough)
image
4. Notice that the `grant_write_permission` fails with the error message `attached deposit is less than the minimum storage balance`

**Expected behavior**
The DevHUB UI should be able to detect the needed attached deposit, and not deposit an amount that is too low and cause this error message.

Megha-Dev-19:
Resolves #629 
Preview [link](https://near.org/megha19.near/widget/app?page=create)

### DIFF
--- a/replacements.dev.json
+++ b/replacements.dev.json
@@ -4,5 +4,6 @@
   "REPL_DEVHUB_CONTRACT": "thomaspreview.testnet",
   "REPL_NEAR": "discom.testnet",
   "REPL_MOB": "eugenethedream",
-  "REPL_EFIZ": "efiz.testnet"
+  "REPL_EFIZ": "efiz.testnet",
+  "REPL_SOCIAL_CONTRACT": "v1.social08.testnet"
 }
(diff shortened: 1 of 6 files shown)


ailisp (approved):
Good fix! Thanks

PR #638, opened 2024-01-13
Regression test cases

### LINKED ISSUE #593: Regression tests for critical use cases
We hit several regression bugs on post create, post reply, community edit etc. These are core functionality and critical use case of DevHub. We should ensure the correctness of them whenever we try to introduce new functionality, fix bugs or refactor.

## Acceptance criteria
Add playwright tests to:
- create every kinds of post
- reply every kinds of post
- create community
- edit community

For each of above operation, observe visually from playwright controlled browser window, and try to be as comprehensive as possible on assert the resulting behavior.  For example, before edit, the submit button should be disabled. After typing something, the submit button should be enabled and the cancel button should work. After success edit, the editor shouldn't appear. And whenever we hit a regression, we should complement these tests to prevent this bug happen again from any future changes.

petersalomonsen:
added docs for the dev container setup with instructions on how to start the test environment from within it
added test for create and edit community

part of implementing #593

### DIFF
--- a/playwright-tests/tests/communities.spec.js
+++ b/playwright-tests/tests/communities.spec.js
@@ -13,33 +13,19 @@ test.describe("Wallet is connected", () => {
   test("should show spawner when user clicks create community", async ({
     page,
   }) => {
-    await page.goto("/devgovgigs.near/widget/app?page=communities");
+    await page.goto("/devhub.near/widget/app?page=communities");
 
-    const createCommunityButtonSelector = 'button:has-text("Create Community")';
+    await page.getByRole("button", { name: " Community" }).click();
 
-    await page.waitForSelector(createCommunityButtonSelector, {
-      state: "visible",
-    });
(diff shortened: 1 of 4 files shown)


Tguntenaar (approved):

playwright-tests/tests/communities.spec.js
```diff
@@ -91,6 +91,52 @@ test.describe("Wallet is connected", () => {
       true
     );
   });
+  test("should create a new community", async ({ page }) => {
+    await page.goto("/devgovgigs.near/widget/app?page=communities");
```
ailisp, line 95: Should this be `/devhub.near/widget/app/?page=communities`?

petersalomonsen: Indeed it should be. I've changed it for the other community tests too now. Actually that also required some changes in the tests. Thank you for finding this.

PR #644, opened 2024-01-21
Test/create-post

### LINKED ISSUE #593: Regression tests for critical use cases

petersalomonsen:
- test for creating an idea post
- test for replying to an idea post in the feed with a comment

resolves #593

### DIFF
--- a/playwright-tests/tests/announcements.spec.js
+++ b/playwright-tests/tests/announcements.spec.js
@@ -92,7 +92,7 @@ test.describe("Admin wallet is connected", () => {
     await page.goto(
       "/devhub.near/widget/app?page=community&handle=devhub-test"
     );
-    const postLocator = page.locator(".post");
+    const postLocator = page.locator(".post").first();
     await postLocator.focus();
   });
 
--- a/playwright-tests/tests/blog.spec.js
+++ b/playwright-tests/tests/blog.spec.js
@@ -4,7 +4,7 @@ test("should load blogs in the sidebar for a given handle", async ({
   page,
 }) => {
   await page.goto(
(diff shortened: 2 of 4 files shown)


playwright-tests/tests/announcements.spec.js
```diff
@@ -92,7 +92,7 @@ test.describe("Admin wallet is connected", () => {
     await page.goto(
       "/devhub.near/widget/app?page=community&handle=devhub-test"
     );
-    const postLocator = page.locator(".post");
+    const postLocator = page.locator(".post").first();
```
ailisp, line 95: Thanks for the fix!

ailisp (approved):
Great reference tests! I suggest to merge it now. Also we're going to supersede post with proposal workflow and near social post soon, so let's not add additional post tests and it will be test/create-proposal. More info about proposals: https://github.com/NEAR-DevHub/neardevhub-bos/milestone/15. Communities are changed much and will still need community tests

PR #670, opened 2024-02-22
fix(discussions): content match before repost on discussions

petersalomonsen:
Check the new discussion posted content with the post on SocialDB, and wait for it to match before reposting on discussions with `blockHeight`. This is to ensure that we don't repost the reference of a previous SocialDB post.

Preview URL: https://near.social/devhub.near/widget/app?page=community&handle=webassemblymusic&tab=discussions

Here's a video for one of the tests, when the content match, then the `create_discussion` confirmation dialog will show. The other test, where the content does not match, is just verifying that the dialog does not show after some seconds.

https://github.com/NEAR-DevHub/neardevhub-bos/assets/9760441/3b63b877-cb58-4939-96f7-dda8841abcdb


frol (approved):
Looks good to me!

PR #673, opened 2024-02-26
updated price on the tx

Tguntenaar:
_Resolves #672_


Tguntenaar:
The test that is failing is not related to this problem so will fix that in a different issue

PR #677, opened 2024-02-27
test(discussions): redirect tx lookup to archival node

petersalomonsen:
Fix for discussions test that is not able to find the example transaction on mainnet RPC, by redirecting the request to archival RPC instead.


Megha-Dev-19 (approved):
LGTM

PR #683, opened 2024-03-04
Proposals bug fixes, notifications and mobile responsiveness

Megha-Dev-19:
Fixed error mentioned in [sheet](https://docs.google.com/spreadsheets/d/1uaKGpIJl_mk1lpeFLAImv5s9XSvsR_s-kRzOKhuljwQ/edit#gid=0)
Notification near.org [PR](https://github.com/near/near-discovery-components/pull/692)
Preview [link](https://near.org/megha19.near/widget/app?page=proposals)


src/devhub/entity/proposal/Proposal.jsx
```diff
@@ -391,21 +395,58 @@ const RadioButton = ({ value, isChecked, label }) => {
...
+      contractName: "${REPL_SOCIAL_CONTRACT}",
+      methodName: "grant_write_permission",
+      args: {
+        predecessor_id: "${REPL_DEVHUB_CONTRACT}",
+        keys: [accountId + "/index/notify"],
+      },
+      gas: Big(10).pow(14),
+      deposit: getDepositAmountForWriteAccess(userStorageDeposit),
+    });
+  }
```
frol, line 448: @Megha-Dev-19 I believe that we don't need to grant permissions anymore: https://github.com/near/near-discovery-components/pull/692#pullrequestreview-1926715790

`devhub/mention` and `devhub/edit` are now posted on behalf of devhub.near contract and thus it does not need to have permissions to post on behalf of the user anymore.

`devhub/reply` and `devhub/like` are now originated from the UI and interacts with SocialDB directly, so users will sign their transaction reaching SocialDB directly, so we don't need to setup permissions there either.

Megha-Dev-19: got it, I will remove it then
so for notifications I will need the editor_id to show the right person profile, rather than devhub.near account, right?
can we pass it in value for notification data?
or should I fetch proposal data using the proposal id?

Tguntenaar (approved):

PR #687, opened 2024-03-05
UX improvements for "Don't ask again"

### LINKED ISSUE #680: Indicator while waiting for post to be submitted and page refreshed ( when "Don't ask again" is enabled )

When "Don't ask again" is enabled, I'm able to submit a post without confirming with the wallet. Still there are some delays, without any indication, that might lead me to think that I need to reload the page.

Here's a video ( turn audio on ) that demonstrates these delays, where you can see there is no indication that I need to wait:

https://github.com/NEAR-DevHub/neardevhub-bos/assets/9760441/a4282efc-6242-4771-9748-ab4cde4ef69b

I would like a toast to show up immediately when I click the submit button, in the delay between the button was clicked and the transaction sending starts. Also I would like a toast to show up when the transaction is completed from the VM, but while still refreshing the page is not completed. This way I would have a clear indication that I don't need to reload the page.

petersalomonsen:
# Preview environment

https://near.org/devgovgigs.petersalomonsen.near/widget/app?page=post&id=3135

# Implementation plan

## Tests
- [X] upgrade to latest near-social-vm (near-bos-webcomponent) in tests
- [X] Set up test scenarios with "Don't ask again" configured ( accesskey to `devhub.near` in localstorage, and widget permission in indexeddb cache ( VM local storage) )
- [X] Handle that the VM is not able to detect the widget when `bos-loader` is used, by modifying RPC responses from `social.near` / `get`
- [X] Modify response for `view_access_key_list` RPC call to avoid wallet redirect
- [X] Modify response for sending transaction, so that the test get a response that the transaction is complete
- [X] Wait for contents to refresh after cache invalidation, and assert that there is an indicator while waiting

## Implementation
- [X] Hide submit/cancel buttons when submitting the transaction
- [ ] Ensure that these buttons should not be removed if there is a confirmation dialog ("Don't ask again" is not enabled), and the user selects cancel. **BLOCKED** Depends on https://github.com/NearSocial/VM/issues/105
- [X] Indicator that automatic refresh is still going on after the transaction is completed, to avoid that the user thinks the page needs to be reloaded
- [X] Clicking "Like" should also prevent double clicking, and also indicate something while refreshing is still going on

# Screen videos ( generated by test runner )

The following video shows the hiding of the submit and cancel button when clicking the submit button, and it also shows a loading indicator all the way until the contents of the page is submitted. Note that it might seem confusing that you don't see the submitted comment after refresh, but this is only the behaviour in the test-runner, where there is no actual transaction submitted.

https://github.com/NEAR-DevHub/neardevhub-bos/assets/9760441/1c50fc75-51c1-43d7-84a4-64cb09323feb

In the next video everything is quite slow, just like in the video from the issue where I pointed out the problem. This fix does not intend to solve the slowness problem that happens when interacting with a long thread like this one, but rather makes sure that there are indicators showing that there are things going on.

https://github.com/NEAR-DevHub/neardevhub-bos/assets/9760441/aec7c722-b0bb-4d75-87c7-079870a149a9

This fix will also introduce a new problem. If "Don't ask again" is not enabled, and you close the transaction confirmation dialog ( https://github.com/NearSocial/VM/issues/105 ), then the loading indicator is still there, and the submit button is still hidden. I haven't found any way to solve this as long as we don't have a callback method for when the transaction confirmation dialog is closed. 

Here's a video showing that problem:

https://github.com/NEAR-DevHub/neardevhub-bos/assets/9760441/8d23a902-e2dd-4817-98d8-27a7ef52a788

Finally here's a video demonstrating how clicking the like button is handled. Here you can see that double-clicks are prevented by hiding the button immediately when it's clicked. Also you can see that the loading indicator is showing until the new like is updated.

https://github.com/NEAR-DevHub/neardevhub-bos/assets/9760441/bee07b9a-2729-440c-a120-7de483398932

fixes #680


petersalomonsen:
A challenge I discovered when working on this is that setting `flags` at https://near.org/flags to `http://127.0.0.1:3030` where there is a `bos-loader` running prevents the transaction confirmation dialog to detect the widget, which is needed by the "Don't ask again" feature.

It looks like this:

image

Because of this the only way to verify the fix is currently by deploying to a preview environment, or by intercepting the RPC calls to retrieve the widget in the Playwright test runner. I'm currently looking into mocking the RPC call to return the widget the local development environment.

petersalomonsen:
Here's a screen video showing the first part of fixing this, which is removing the submit and cancel buttons from the view when submitting the transaction. Also note the don't ask again toast in the bottom right.

https://github.com/NEAR-DevHub/neardevhub-bos/assets/9760441/ca736439-c3e4-421a-bae2-5c024f2230df

The next step is to ensure that these buttons should not be removed if there is a confirmation dialog, and the user selects cancel. I will add this to the acceptance criteria.

frol:
@petersalomonsen Would it be easy to disable the buttons and add loader inside the button instead of completely removing the button? This is what I mean:

image

Regarding the blocker - the only solution I can think of is to have 15 seconds timeout for this loader state and also have an additional check if the post is already recorded on chain - this complicates the logic quite a bit, so I would only implement the timeout part if possible.

petersalomonsen:
@frol Having the loading spinner inside the button, and disabling it, was an easy change. The preview environment is updated with this. I did set the spinner on the left side of the button as it is in the Bootstrap docs ( https://getbootstrap.com/docs/4.2/components/spinners/ ).

I haven't implemented the timeout yet. It feels rather hacky. I believe your suggestion could work, though for long threads even 15 seconds might not be enough, which you also can see in the video with the long thread. I'm not sure how critical this is, but I feel the appropriate would be to start implementing support for a callback in the VM when the confirmation dialog is closed.

frol (approved):

PR #691, opened 2024-03-14
test setup for discussions

petersalomonsen:


playwright-tests/tests/community.spec.js
```diff
@@ -43,6 +43,7 @@ test.describe("Wallet is connected", () => {
   test("should allow connected user to post from community page", async ({
     page,
   }) => {
+    test.setTimeout(60000);
```
Megha-Dev-19, line 46: just to confirm, have you added it because of the RPC issue?

playwright-tests/tests/discussions.spec.js
```diff
@@ -142,3 +154,156 @@ test.describe("Wallet is connected", () => {
...
+      page
+    );
+
+    let discussion_created = false;
+    await mockSocialIndexResponses(page, ({ requestPostData, json }) => {
+      if (
+        requestPostData.action === "repost" &&
+        requestPostData.options?.accountId?.[0] === communitydiscussionaccount
+      ) {
+        console.log("Returning discussions from index", discussion_created);
```
Megha-Dev-19, line 178: do we need logs inside test?

playwright-tests/tests/funding.spec.js
```diff
@@ -11,24 +15,35 @@ test.describe("Wallet is connected", () => {
...
+    await pauseIfVideoRecording(page);
 
     const tagsInput = page.locator(".rbt-input-multi");
+    await tagsInput.scrollIntoViewIfNeeded();
     await tagsInput.focus();
     await tagsInput.pressSequentially("paid-cont", { delay: 100 });
     await tagsInput.press("Tab");
     await tagsInput.pressSequentially("developer-da", { delay: 100 });
     await tagsInput.press("Tab");
+    await pauseIfVideoRecording(page);
```
Megha-Dev-19, line 46: I don't understand, why do we check it after every step? shouldn't we check only once if video exists.

Megha-Dev-19 (approved):
LGTM, Thanks @petersalomonsen for adding the tests for proposals.

playwright-tests/tests/community.spec.js
petersalomonsen: yes I observed that the tests where I've increased the timeout often fails because of timeout. The default is 30000, so increasing it to 60000 should give it more time to finish.

playwright-tests/tests/funding.spec.js
petersalomonsen: The `pauseIfVideoRecording` method checks if you have set `video` to `on` in your playwright.config.js here:

image

If video recording is enabled, then the `pauseIfVideoRecording` method will just insert a 0.5 second delay before moving on with the next step. This is just so that the video should be a bit easier to watch, and not running too fast. In the pipeline we want as few delays as possible, and there video recording should be set to `off`, and there will be no delay.

playwright-tests/tests/discussions.spec.js
petersalomonsen: No, not mandatory. The reason for logging here was to track where we are intercepting calls to the social index API. I can remove it if it's too annoying.

Megha-Dev-19: No it’s fine 👍

PR #694, opened 2024-03-19
Feature/690 filter search proposals

Tguntenaar:
_Resolves #690_
[PREVIEW](https://near.org/thomasguntenaar.near/widget/app?page=proposals)

This PR fixes the search / filters on the new proposals page. The only thing that still has to happen is the sorting functionality. More specifically sorting on comments. However because the launch is this week I discussed with @ailisp to merge the finished part already.

I'm currently still working on tracking the amount of comments per proposal. If we want to track the number of comments in the indexer, I think the easiest would be to add a function to devhub.near that is triggered together with the social.set function. That way it is easy for the indexer to track it.

Another solution might be to query that from social db indexer, which you can get all comments of a given post? Not sure we can query another indexer from this indexer's indexingLogic. So that is reason I wanted to merge this before trying that.

### DIFF
--- a/src/devhub/entity/proposal/Feed.jsx
+++ b/src/devhub/entity/proposal/Feed.jsx
@@ -185,8 +185,6 @@ const FeedPage = () => {
       timeline
     }
     ${queryName}_aggregate(
-      offset: $offset
-      limit: $limit
       order_by: {proposal_id: desc}
       where: $where
     )  {
--- a/src/devhub/entity/proposal/Feed.jsx
+++ b/src/devhub/entity/proposal/Feed.jsx
@@ -61,7 +61,6 @@ const Heading = styled.div`
 const FeedItem = ({ proposal }) => {
   const accountId = proposal.author_id;
   const profile = Social.get(`${accountId}/profile/**`, "final");
-  // FIXME: social_db_post_block_height is only know in proposal not the body. Which means it won't be indexed.
   // We will have to get the proposal from the contract to get the block height.
   const blockHeight = parseInt(proposal.social_db_post_block_height);
   const item = {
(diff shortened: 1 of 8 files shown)


ailisp:
Great implementation! Every search works well! Let's address Megha's concern (activity feed search UI is probably a good reference of these)

> So that is reason I wanted to merge this before trying that.

I agree

Tguntenaar:
@Megha-Dev-19  @ailisp thank you  for the review! I fixed the bugs you encountered.

Megha-Dev-19:
LGTM, Pre approving:
- can we add none or clear option to author dropdown as well
- the filters positioning looks different from figma file
In feed we currently fetch all proposals (which isn't efficient), in design we have load more, we add it in other PR also, just wanted to let you know.

src/devhub/entity/proposal/Feed.jsx
```diff
@@ -142,47 +143,218 @@ const FeedItem = ({ proposal }) => {
...
+      order_by: {proposal_id: asc}
+      where: $where
+    )  {
+      aggregate {
+        count
+      }
+    }
+  }`;
+  function fetchGraphQL(operationsDoc, operationName, variables) {
+    console.log({ variables });
```
Megha-Dev-19, line 198: can we remove the log?

src/devhub/entity/proposal/Feed.jsx
```diff
@@ -142,47 +143,218 @@ const FeedItem = ({ proposal }) => {
...
+    fetchGraphQL(query, "GetLatestSnapshot", variables).then(async (result) => {
+      if (result.status === 200) {
+        if (result.body.data) {
+          const data =
+            result.body.data
+              .thomasguntenaar_near_devhub_proposals_november_proposals_with_latest_snapshot;
+          const totalResult =
+            result.body.data
+              .thomasguntenaar_near_devhub_proposals_november_proposals_with_latest_snapshot_aggregate;
+          console.log({ data, count: totalResult.aggregate.count });
```
Megha-Dev-19, line 266: there are quite a few logs

Megha-Dev-19 (approved):

PR #699, opened 2024-03-22
fix indexer recently stopped due to queryapi bug

ailisp:
Fix feed not loading for recent posts after Mar. 17 and recent blog post not showing up. Root cause is a query api bug that stopped the working indexer, and still under investigation of query api team. This fix address the problem for now, but if the bug is not fixed in query api side, future posts/blogs can still not loading in some indefinite future time


ailisp:
As this is a straightforward change and I tested in local preview, I'll just merge it.

PR #713, opened 2024-03-26
fix(proposals): blur 'get verified' drop-down

petersalomonsen:
test case for reproducing not able to blur 'get verified' drop-down


Megha-Dev-19 (approved):
LGTM

PR #717, opened 2024-03-30
feat(communities): announcements don't ask again

petersalomonsen:
preview environment:

https://near.social/devgovgigs.petersalomonsen.near/widget/app

show loading indicator for announcements when submitting, and hide when transaction is finished.

Note that the loading indicator stops spinning before the feed is updated. It is possible to make it spin until the feed is updated if we drop the fallback to `Social.index`, but I'm not sure how critical this is.

https://github.com/NEAR-DevHub/neardevhub-bos/assets/9760441/6d8fd1c7-ec4d-4c7d-9f6b-bd3e0ffa4530

part of resolving #687


petersalomonsen:
@Megha-Dev-19 The reason for the proposals lagging is the effects having the dependency to getting all the proposals data. I've fixed one of them, but we need to find another solution for the linked proposals list.

Tguntenaar (approved):

PR #736, opened 2024-04-11
Fix proposal timestamp and KYC dropdown action

Megha-Dev-19:
Preview [link](https://near.org/megha19.near/widget/app?page=proposasl)


Tguntenaar (approved):
lgtm, we might want to add a test that it shows the right date?

PR #739, opened 2024-04-12
View feed as community.devhub.near as anonymous user

Tguntenaar:
_Resolves #737_

[Preview](https://near.org/thomasguntenaar.near/widget/app?page=announcements)


Megha-Dev-19 (approved):
LGTM

frol (approved):
As we discussed in the chat, let's merge it as is for now, but find a better solution to avoid the containerization (60% scrollable view) since it is not a pleasant user experience

Tguntenaar (commented):
Ready to merge

PR #740, opened 2024-04-12
Proposal form lag fix

Megha-Dev-19:
Preview [link](https://near.org/megha19.near/widget/app?page=create-proposal)


petersalomonsen (approved):
So no need for the pressSequentially delay anymore? great!

playwright-tests/tests/proposals.spec.js
```diff
@@ -194,34 +189,12 @@ test.describe("Wallet is connected", () => {
...
-      {
-        delay: delay_milliseconds_between_keypress_when_typing,
-      }
+      "The test proposal description. And mentioning @petersalomonsen.near. Also mentioning @megha19.near"
     );
 
     await pauseIfVideoRecording(page);
 
-    await page.frameLocator("iframe").getByText("petersalomonsen.near").click();
-
```
petersalomonsen, line 206: @Megha-Dev-19 you may want to keep this, because this is testing some properties of the markdown editor, This was the fix we made for pressing backspace leaving only the @, and then start typing again ( when the dropdown used to not show up again ). We may want to have such tests

petersalomonsen (changes requested):

petersalomonsen (approved):

PR #764, opened 2024-04-24
Fix compose comment for proposals

Megha-Dev-19:
Preview [link](https://near.org/megha19.near/widget/app?page=proposals)


petersalomonsen:
Ready @Megha-Dev-19

petersalomonsen:
@ori-near @frol This Pull Request by @Megha-Dev-19 should address many of the commenting issues on proposals that we discussed in the call yesterday. I did a small adjustment to clear the editor when the submission transaction is finished, and also added "Don't ask again" configuration to the tests ( which is slightly different when saving to socialdb compared to calling another smart contract like devhub.near ).

Here is a video of what the automatic test now does. As you can see it creates the comment and submit. There is no "blue toast" from the VM like when calling another contract, but there's the disabled submit button with a spinner. In the end there's a also a confirmation toast. What this test does not do is to show the actual submitted comment in the end, and if we want this too, it is possible by mocking the results of the subscribed social.index.

https://github.com/NEAR-DevHub/neardevhub-bos/assets/9760441/e9777a28-548f-4603-81d6-3710a682efab

PR #784, opened 2024-05-21
Removed top level fixed positioning

### LINKED ISSUE #782: Near org banner hides navbar
1) Go to [feed page](https://near.org/devhub.near/widget/app?page=proposals)
2) Navbar is hidden
image

Check for all instances
- [ ]  Devhub proposals
- [ ]  Infrastructure committee -> Rfps and proposals
- [ ]  Events committee

Megha-Dev-19:
Resolves #782 
Resolves [#1163](https://github.com/near/near-discovery/issues/1163)

https://github.com/NEAR-DevHub/neardevhub-bos/assets/100185149/ddba5763-77dc-4121-a8ef-ac3c7f1c6c7b

### DIFF
--- a/src/devhub/components/templates/AppLayout.jsx
+++ b/src/devhub/components/templates/AppLayout.jsx
@@ -1,5 +1,9 @@
+const data = fetch(`https://httpbin.org/headers`);
+const gatewayURL = data?.body?.headers?.Origin ?? "";
+
+// we need fixed positioning for near social and not for org
 const Theme = styled.div`
-  position: fixed;
+  position: ${gatewayURL.includes("near.org") ? "inherit" : "fixed"};
   inset: 73px 0px 0px;
   width: 100%;
   display: flex;
--- a/src/devhub/components/templates/AppLayout.jsx
+++ b/src/devhub/components/templates/AppLayout.jsx
@@ -2,13 +2,20 @@ const data = fetch(`https://httpbin.org/headers`);
 const gatewayURL = data?.body?.headers?.Origin ?? "";
 
 // we need fixed positioning for near social and not for org
+const ParentContainer = gatewayURL.includes("near.org")
+  ? styled.div`
+      width: 100%;
+    `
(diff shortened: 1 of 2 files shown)


Megha-Dev-19:
@Tguntenaar should we merge it and then create a PR from your `main` to `events-committee`, I see it has some left commits, or should I add a PR right now with this change only?
image

petersalomonsen (approved):
Looking good @Megha-Dev-19

Megha-Dev-19:
Thanks @petersalomonsen for the great solution to fetch the header.

src/devhub/components/templates/AppLayout.jsx
```diff
@@ -1,5 +1,9 @@
+const data = fetch(`https://httpbin.org/headers`);
+const gatewayURL = data?.body?.headers?.Origin ?? "";
+
+// we need fixed positioning for near social and not for org
 const Theme = styled.div`
-  position: fixed;
+  position: ${gatewayURL.includes("near.org") ? "inherit" : "fixed"};
```
petersalomonsen, line 6: using an expression inside the`styled.div` leads to that the transaction modal stops working (which the tests detect and then fails). So instead you have to do something like this:

image

petersalomonsen (changes requested):

src/devhub/components/templates/AppLayout.jsx
Megha-Dev-19: I have made the update

petersalomonsen (approved):

Tguntenaar:
> @Tguntenaar should we merge it and then create a PR from your `main` to `events-committee`, I see it has some left commits, or should I add a PR right now with this change only? 

What do you mean with left commits? I think merging `main` to `events-committee` should be done carefully. Might be better to cherry pick some commits. Since `main` contains code that has been deleted in the events-commitee. For example, the footer tried to sneak back in.

PR #797, opened 2024-05-30
Fix proposals feed filter

### LINKED ISSUE #781: Issues with search and filtering of feed
1) Go to [feed page](https://near.org/devhub.near/widget/app?page=proposals)
2) Filtering and search works sometimes, otherwise it keeps the data same

Check for all instances
- [x] Devhub proposals
- [x] Infrastructure committee -> Rfps and proposals
- [x] Events committee

Megha-Dev-19:
Resolves #781 
https://drive.google.com/file/d/1QNVfxx79L0dTvriyLKbprjZRvQFe3awv/view?usp=sharing
[video.webm](https://github.com/NEAR-DevHub/neardevhub-bos/assets/100185149/bb73ab84-7e91-4bd9-8583-667667f6bcf8)
[video.webm](https://github.com/NEAR-DevHub/neardevhub-bos/assets/100185149/e913ccd0-075e-4ccc-907b-892a3dac8c58)
[video.webm](https://github.com/NEAR-DevHub/neardevhub-bos/assets/100185149/a883053a-300f-4971-8751-b6f3dd8c58a1)
[video.webm](https://github.com/NEAR-DevHub/neardevhub-bos/assets/100185149/55cc31bb-a29c-4985-8025-0cdfedb9e2e2)


petersalomonsen (changes requested):
For tests here, we should just use the live content and click and change filters like you did in the video, and check that we get results according to it.

Megha-Dev-19:
@petersalomonsen I have added some basic test code, but I don't understand how do we check the results? should we check it using the response data or should we do it from the UI? like all the category tags should have specific input? 
and how can we achieve that?

playwright-tests/tests/proposals.spec.js
```diff
@@ -455,4 +455,45 @@ test.describe("Wallet is connected", () => {
...
+    await page.getByRole("button", { name: "Stage" }).click();
+    await page.getByRole("list").getByText("Funded").click();
+    await expect(
+      page.getByRole("button", { name: "Stage : Funded" })
+    ).toBeVisible();
+    test.setTimeout(4000);
+  });
+
+  test("should filter proposals by author", async ({ page }) => {
+    test.setTimeout(120000);
```
petersalomonsen, line 482: Do we need 2 minutes timeout here?

playwright-tests/tests/proposals.spec.js
```diff
@@ -455,4 +455,45 @@ test.describe("Wallet is connected", () => {
...
+    await page.goto("/devhub.near/widget/app?page=proposals");
+    await page.getByRole("button", { name: "Author" }).click();
+    await page.getByRole("list").getByText("megha19.near").click();
+    await expect(
+      page.getByRole("button", { name: "Author : megha19.near" })
+    ).toBeVisible();
+    test.setTimeout(4000);
```
petersalomonsen, line 489: why setting timeout here?

playwright-tests/tests/proposals.spec.js
Megha-Dev-19: we have this timeout before every test, which I believe is to load the page

petersalomonsen:
> @petersalomonsen I have added some basic test code, but I don't understand how do we check the results? should we check it using the response data or should we do it from the UI? like all the category tags should have specific input? and how can we achieve that?

We should check the results in the UI. So that when filtering, we should also add expectations that the results visible to the user matches the filter. Checking just the response data will not be sufficient to verify that this is actually what the user see, so we need to check what is visible. Also checking that it is not lagging, by checking the time it takes to get a result is relevant.

playwright-tests/tests/proposals.spec.js
Megha-Dev-19: to wait for new filtered data to be fetched.

petersalomonsen: `test.setTImeout` is for setting how long the test is allowed to run ( default is 30 seconds ). If you need to wait for something to appear on the page, you can normally just use `await expect(await page.locator(".some-class-name")).toHaveText("blabla")`, like you can see here: https://github.com/NEAR-DevHub/neardevhub-bos/pull/802/files#diff-1765f31365d3e8010bc399faa0e2b805b801dbf1cb84c5abfe7e06184f57fa28R329

The default setting is that it waits for up to 5  seconds, but you can also add more as a setting to the `page.locator` method.

Megha-Dev-19:
@petersalomonsen I have updated the tests and added their videos too, can you please review.

Tguntenaar (approved):
LG

PR #802, opened 2024-06-02
Edit proposal timeline with "don't ask again" enabled

### LINKED ISSUE #800: Add confirmations after editing proposal timeline with "Don't ask again" enabled
When approving a proposal with "Don't ask again" enabled there is no confirmation of the change.

The blue "sending transaction" toast is showing, but after the only indicator is the change of the status badge under the proposal title.

There should be a confirmation popup confirming that the change succeeded.

After that you also have to reload the page to see history entries for the change of the proposal state

image

*No entries are added here without reloading the page*

The log should be updated with entries without having to reload.

A confirmation toast should display after the change is effected, and the log should be updated with entries about changing of the proposal state.

# Tasks

- [x] add test case for approving a proposal with don't ask again enabled
- [x] Show confirmation toast after the change submission is successful
- [x] Update log entries without having to reload the page

petersalomonsen:
- add test case for approving a proposal with don't ask again enabled
- update log without having to refresh after saving changed timeline status
- confirmation toast when editing timeline update transaction is finished

https://github.com/NEAR-DevHub/neardevhub-bos/assets/9760441/f57b70d7-bdca-437b-a177-b27aefc54b44

fixes #800

### DIFF
--- a/playwright-tests/storage-states/wallet-connected-with-devhub-moderator-access-key.json
+++ b/playwright-tests/storage-states/wallet-connected-with-devhub-moderator-access-key.json
@@ -1,0 +1,51 @@
+{
+  "cookies": [],
+  "origins": [
+    {
+      "origin": "http://localhost:8080",
+      "localStorage": [
+        {
+          "name": "near-wallet-selector:selectedWalletId",
+          "value": "\"my-near-wallet\""
+        },
+        {
+          "name": "near_app_wallet_auth_key",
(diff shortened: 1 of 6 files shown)


petersalomonsen:
please review @Megha-Dev-19 , and also check the video @ori-near if it match your expectations

Megha-Dev-19 (approved):
LGTM

PR #832, opened 2024-06-07
Fix fractal status when API is down

### LINKED ISSUE #810: Handle Fractal API error
Fractal API can return error, so instead of showing "Not verified", we should show network error state

Instances
- [x] Devhub
- [x]  IC
- [ ] Events Committee

Megha-Dev-19:
Resolves #810 
image

This will fix it for devhub and IC


Tguntenaar (approved):
LG

PR #836, opened 2024-06-08
Generalise custom notifications

### LINKED ISSUE #803: Generalise custom notification
Currently we use `devhub/like, devhub/edit, /mention, /reply` for proposal notifications, but since now we have multiple instances @frol suggested that we should use `proposal/like`, `rfp/like` such endpoints.
We will need to update our codebase, near.org and mob's widget (for near.social) to support it.

Megha-Dev-19:
Resolves #803 
Instead of using `devhub/mention` we will use `proposal/mention` and `rfp/mention`, so it works for all instances and `widgetAccountId` to navigate user to correct URL

### DIFF
--- a/src/devhub/entity/proposal/ComposeComment.jsx
+++ b/src/devhub/entity/proposal/ComposeComment.jsx
@@ -154,7 +154,7 @@ function composeData() {
         type: "proposal/reply",
         item,
         proposal: proposalId,
-        widgetAccountId: "REPL_DEVHUB",
+        widgetAccountId: "${REPL_DEVHUB}",
       },
     });
   }
--- a/src/devhub/entity/proposal/ComposeComment.jsx
+++ b/src/devhub/entity/proposal/ComposeComment.jsx
@@ -151,9 +151,10 @@ function composeData() {
     notifications.push({
       key: notifyAccountId,
       value: {
-        type: "devhub/reply",
+        type: "proposal/reply",
         item,
         proposal: proposalId,
+        widgetAccountId: "REPL_DEVHUB",
       },
     });
   }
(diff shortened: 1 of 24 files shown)


src/devhub/entity/proposal/ComposeComment.jsx
```diff
@@ -151,9 +151,10 @@ function composeData() {
     notifications.push({
       key: notifyAccountId,
       value: {
-        type: "devhub/reply",
+        type: "proposal/reply",
         item,
         proposal: proposalId,
+        widgetAccountId: "REPL_DEVHUB",
```
Tguntenaar, line 157: Is it not `${REPL_DEVHUB}`?

src/devhub/notification/Left.jsx
```diff
@@ -16,18 +17,33 @@ return props.type ? (
...
-        },
-      })}
-    >
-      DevHub proposal
-    </a>
+    {props.proposal ? (
+      <a
+        className="fw-bold text-muted"
+        href={href({
+          widgetSrc: `${widgetAccountId}/widget/app`,
```
Tguntenaar, line 24: If the widgetAccountId is a static string I wouldn't it in a variable here. It will not be replaced during runtime.

Tguntenaar (changes requested):

src/devhub/notification/Left.jsx
Megha-Dev-19: this is not for our widgets, we will use it inside mob's notification widget, it will be passed from props.

src/devhub/entity/proposal/ComposeComment.jsx
Megha-Dev-19: good catch, I will update it

Tguntenaar (approved):

PR #845, opened 2024-06-19
Feature/746 relevant mentions

Tguntenaar:
_Resolves #746_

@ori-near I've added the supervisor and requested sponsor as relevant users to the acceptance criteria. The last ac will never happen since there will always be an author and sponsor. There is currently a limit of 5 users + any relevant users that are suggested when a users types `@`. I could add a scroll in the suggestion box if you want to show all devhub users.

PS Instead of all devhub users we could show all proposal authors? This is a list that is available via a contract call. Are there more users we count as devhub users but won't be included in this list.

Acceptance Criteria:
 - [x] When a user types "@" while composing a comment on a proposal, the dropdown prioritizes users relevant to the discussion
   - [x] First: The proposal's author
   - [x] Second supervisor
   - [x] Third requested sponsor
   - [x] Fourth: Users who have recently commented on the proposal
 - [ ] If there are no relevant users (e.g. the user is the first comment), the dropdown displays all DevHub users in alphabetical order

Screenshot 2024-06-19 at 12 21 07
Screenshot 2024-06-19 at 12 21 16

[video.webm](https://github.com/NEAR-DevHub/neardevhub-bos/assets/28901891/9d1ebc99-c5f5-4f29-911e-9b38aad51afc)

### DIFF
--- a/instances/devhub.near/widget/devhub/components/molecule/SimpleMDE.jsx
+++ b/instances/devhub.near/widget/devhub/components/molecule/SimpleMDE.jsx
@@ -11,6 +11,7 @@ const onChangeKeyup = props.onChangeKeyup ?? (() => {}); // in case where we wan
 const height = props.height ?? "390";
 const className = props.className ?? "w-100";
 const embeddCSS = props.embeddCSS;
+const sortedRelevantUsers = props.sortedRelevantUsers || [];
 
 State.init({
   iframeHeight: height,
@@ -136,10 +137,12 @@ let showAccountAutoComplete = ${showAccountAutoComplete};
 let showProposalIdAutoComplete = ${showProposalIdAutoComplete};
 
 function getSuggestedAccounts(term) {
-  let results = [];
+  let results = [${sortedRelevantUsers
+    .map((u) => "{accountId:'" + u + "', score: 100}")
+    .join(",")}];
 
   term = (term || "").replace(/\W/g, "").toLowerCase();
-  const limit = 5;
+  const limit = ${5 + sortedRelevantUsers.length};
 
   const profiles = Object.entries(profilesData);
 
--- a/playwright-tests/tests/proposal/proposals.spec.js
+++ b/playwright-tests/tests/proposal/proposals.spec.js
@@ -456,6 +456,58 @@ test.describe("Wallet is connected", () => {
     await pauseIfVideoRecording(page);
   });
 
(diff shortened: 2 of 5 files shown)


instances/devhub.near/widget/devhub/components/molecule/SimpleMDE.jsx
```diff
@@ -139,7 +140,13 @@ function getSuggestedAccounts(term) {
   let results = [];
 
   term = (term || "").replace(/\W/g, "").toLowerCase();
-  const limit = 5;
+  let limit = 5;
+  if (term.length < 2) {
```
Megha-Dev-19, line 144: why do we have this term length check?

playwright-tests/tests/proposal/proposals.spec.js
```diff
@@ -456,6 +456,60 @@ test.describe("Wallet is connected", () => {
...
+      "neardevdao.near", //  requested_sponsor,
+      // "geforcy.near", // comment author,
+    ];
+    let mentionSuggestions = [];
+    for (let i = 0; i < liLocators.length; i++) {
+      const text = await liLocators[i].innerText();
+      mentionSuggestions.push(text);
+    }
+    // TODO should be 0,4 including geforcy.near but somehow mob.near shows up.
+    // When I manually test, it shows the correct 4 users
```
Megha-Dev-19, line 508: so do you want to fix it before merging, so can we merge it?

Megha-Dev-19 (approved):
LGTM, left few comments.

instances/devhub.near/widget/devhub/components/molecule/SimpleMDE.jsx
Tguntenaar: If a user doesn't type anything, the most relevant users will appear first. However, if a user types a specific name, the list should display the matching name instead of keeping the most relevant users at the top, followed by the match.

playwright-tests/tests/proposal/proposals.spec.js
Tguntenaar: I will change it before we merge. I have an idea on why it happens, we should mock the Social.index call.

Tguntenaar (commented):
The reason turned out to be the cached content box because of the React Hook useMemo.

I also added that it won't recommend yourself.

PR #854, opened 2024-06-23
Update terms and condition

### LINKED ISSUE #847: Add new T&C log to Events & DevHub proposal create timeline
Next week, we will update the T&C for DevHub & Events, which will eliminate the need for separate agreements and expedite the funding distribution process. However, to support this, our legal team must have a public on-chain record when recipients select the T&C.

# User Story
As the NF legal & finance teams, we need a public on-chain record associated with each proposal so that we can ensure compliance and process payments for each recipient.

# Acceptance Criteria

## For DevHub & Events portals:

- [x] Contract Side: When a user creates a proposal, we must have an explicit accept condition functionality with a timestamp
- [x] UI: Add a log event to the proposal timeline UI after the proposal is created that shows that the user accepted the T&C. 
  - [x] Copy: `[user.near] agreed to the terms & conditions ･ Jun 18, 2024 at 04:38 PM UTC`
- [x] Update the link to the T&C https://drive.google.com/file/d/1ikNsAoc14B1Z94YlOWONPJ1OGBKLtxOk/view?usp=share_link

## For Infrastructure Committee Portal
- [x] Remove T&C reference from both proposal/RFPs. Keep code of conduct reference.

Megha-Dev-19:
Resolves #847 
TODO:

- [x] Deploy changes on contract
- [ ] I will have to change the blockheight once the terms and condition widget is deployed

### DIFF
--- a/instances/devhub.near/aliases.mainnet.json
+++ b/instances/devhub.near/aliases.mainnet.json
@@ -7,5 +7,6 @@
   "REPL_EFIZ": "efiz.near",
   "REPL_DEVS": "devs.near",
   "REPL_SOCIAL_CONTRACT": "social.near",
-  "REPL_RPC_URL": "https://rpc.mainnet.near.org"
+  "REPL_RPC_URL": "https://rpc.mainnet.near.org",
+  "REPL_TERMS_AND_CONDITION_BLOCKHEIGHT": ""
 }
--- a/instances/devhub.near/widget/devhub/entity/proposal/CommentsAndLogs.jsx
+++ b/instances/devhub.near/widget/devhub/entity/proposal/CommentsAndLogs.jsx
@@ -97,6 +97,22 @@ function sortTimelineAndComments() {
           ...getDifferentKeysWithValues(startingPoint, item),
         };
       });
+
+    // add log for accepting terms and condition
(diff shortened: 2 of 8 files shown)


Megha-Dev-19:
@Tguntenaar, we have to add these changes to events as well.

petersalomonsen (approved):

petersalomonsen:
should we put some effort into figuring out why the blog tests are so flaky @Tguntenaar ?

petersalomonsen:
> @petersalomonsen yes we should, I'm working on the blog search performance / a bug that ori had found. Let's see what happens when we re-run this gh action one more time.
> 
> I thought it was RPC related but that should not be the case anymore with your rpc cache enhancements right?

Don't think it is RPC related anymore. It is more probably performance related, laggy components etc. I can take a look at it too, unless you've already found some clues to follow?

petersalomonsen:
@Tguntenaar Back here: https://github.com/NEAR-DevHub/neardevhub-bos/pull/784/commits/3d6e5d84872f845fb979c8e50886ce7ec94434d2, we had tests failing because of using variables inside styled divs. Any chance that this could be the case here too? I see flashes of the same error message as back then which is "VM is dead".

PR #857, opened 2024-06-26
Add infrastructure-committee

Megha-Dev-19:
- [x] Fix tests
- [x] Add script to deploy to infrastructure account (dry run: there will be changes in dry-run since earlier we use to import `common.jsx` instead of replacements)
- [x] Thoroughly test the frontend manually

Need to add following keys to github vars:
- NEAR_INFRA_SOCIAL_ACCOUNT_ID
- NEAR_INFRA_SOCIAL_ACCOUNT_PUBLIC_KEY
- NEAR_INFRA_SOCIAL_ACCOUNT_PRIVATE_KEY
- NEAR_INFRA_SOCIAL_DEPLOY_ID


petersalomonsen:
So as we discussed in the call, we will need:

- A separate copy of the CI workflow targeting the infrastructure committee instance ( running the tests that are relevant )
- A deployment action that does diff ( dry-run ) when not on the main branch
- We also need to include the infrastructure committee test folder in the CI workflow for Infrastructure committee

playwright-tests/tests/infrastructure/rfp.spec.js
```diff
@@ -0,0 +1,309 @@
...
+        console.log(JSON.stringify(originalResult, null, 1));
+        originalResult.snapshot.timeline.status = "ACCEPTING_SUBMISSIONS";
+
+        return originalResult;
+      },
+    });
+
+    await page.goto("/infrastructure-committee.near/widget/app?page=rfp&id=0");
+    await page.getByRole("button", { name: "Edit" }).click();
+    await page.locator(".badge .bi-trash3-fill").click({ timeout: 1000 });
```
petersalomonsen, line 181: This timeout of `1000` ms might be a bit to little. Just leave it to the default which is 5s.

petersalomonsen (approved):
Looks good @Megha-Dev-19 , only minor comments from me

The cancel RFP test is reported as flaky in the test run. Any chance to investigate that to reduce introduction of more flaky tests?

Also there's the blog fix for variables in styled divs in the main branch, so it would be good to run the test suite on top of that as well.

PR #858, opened 2024-07-03
share link button for copying url, share proposal with clean URL

### LINKED ISSUE #851: Use "Clean URLs" when sharing ilnks
Now that we have the web4 gateway with the to create "Clean URLs" like https://devhub.near.page/proposal/127, we should also make sure that the share link buttons generate such links.

Like the buttons here for sharing on twitter and email, they should both point to the https://devhub.near.page and use a clean URL without a query string.

image

Instead of the "share email", we should rather have a "Copy to clipboard" so that it is easy to paste the link anywhere.

petersalomonsen:
- add a share link button for copying url
- share proposal using the new share link button with a clean URL ( e.g. https://devhub.near.page/proposal/22 )
- add test for "Share to X" button
- Remove share to Email button
-  Get local widgets from bos loader via the local RPC proxy, so that there is no need to use a redirect-map with the VM anymore. Also since redirect maps does not works with "don't ask again", this approach eliminates the need for using `modifySocialNearGetRPCResponsesInsteadOfGettingWidgetsFromBOSLoader` in "don't ask again" tests ( since this is now moved to the RPC proxy ). 

https://github.com/NEAR-DevHub/neardevhub-bos/assets/9760441/55de1155-2029-4154-a8b8-56cbeb736b6a

fixes #851

### DIFF
--- a/instances/devhub.near/widget/devhub/components/molecule/ShareLinkButton.jsx
+++ b/instances/devhub.near/widget/devhub/components/molecule/ShareLinkButton.jsx
@@ -1,4 +1,12 @@
+const postType = props.postType ?? "post";
 const externalLink = props.url;
+
+const clickbaitPrompt =
+  props.clickbaitPrompt ??
+  `Check out this ${postType} on @NEARProtocol\n#NEAR #BOS\n${externalLink}`;
+
+const xUrl = new URL("https://x.com/intent/post");
+xUrl.searchParams.set("text", clickbaitPrompt);
 
 const Button = styled.button`
   border: 0;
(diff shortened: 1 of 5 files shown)


petersalomonsen:
@ori-near see the video showing the new "Copy link" button

Tguntenaar (approved):
That is very cool! The rpc proxy is a way better way than multiple playwright servers indeed.

ori-near:
Thank you @petersalomonsen – this is a GREAT change. BTW, is it possible to also introduce the clean URLs on comments URL button? Right now they still point to the the long URL.

For example, the share URL for the below comment:
Screenshot 2024-07-09 at 10 29 26 AM

It currently points to the long URL:
https://near.social/devhub.near/widget/app?page=proposal&id=146&accountId=gagdiez.near&blockHeight=123000964

petersalomonsen:
@ori-near created issue #867 for following up on the comments.

PR #862, opened 2024-07-08
Show latest developer-dao blogs

Tguntenaar:
This PR fixes the following:

- Ori pointed out that the blog link in the navigation bar still pointed to blog v1 instead of v2.
- Blogv2 page was still showing this `blog isn't configured yet` because no communityAddonId was being passed. 
- The provider referenced a function that wasn't defined in that file.

All these bugs are fixed with this PR.


Megha-Dev-19 (approved):
LGTM

PR #864, opened 2024-07-08
Show categories blog page

Tguntenaar:
If you go to [the blog page](https://near.social/devhub.near/widget/app?page=blogv2) the categories are hidden.

This PR:
- Displays the right categories
- Add categorie filter
- Removes some dead code

### DIFF
--- a/instances/devhub.near/widget/devhub/page/blogv2.jsx
+++ b/instances/devhub.near/widget/devhub/page/blogv2.jsx
@@ -1,9 +1,10 @@
 const { id, community } = props;
 
-const { getAccountCommunityPermissions } = VM.require(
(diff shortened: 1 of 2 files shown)


petersalomonsen:
Regarding the new features mentioned in the description, from reading the code, it is not clear to me how "Displays the right categories" and "Add category filter" are implemented. Would it be possible to write a test that demonstrates this?

My guess is that the config you pass on to the other components is what enables these features, but it would be great with a test that expects to see categories and a category filter on the blog page.

Tguntenaar:
Hi @petersalomonsen, I've added a small test that checks wether or not the categories are visible on the main blog page.

petersalomonsen (approved):

PR #902, opened 2024-07-24
Added release for instances

Megha-Dev-19:


petersalomonsen (commented):
Since we now have the protection rules for the prod environments, I suggest moving the dry-run to a separate file with environments in the matrix that does not have protection rules. These environments can be prefixed with `dry-run` and only contain the target account id, since no keys are required for `bos component diff`

PR #965, opened 2024-10-22
Replace near query api endpoint to break tests

Tguntenaar:
WIP

Replaced the nearqueryapi endpoint in the devhub instance to see where the tests would break.


Tguntenaar:
Closing because of duplicate PR https://github.com/NEAR-DevHub/neardevhub-bos/pull/982

PR #993, opened 2024-11-27
Revert "Use cache-api instead of QueryAPI for Proposals and Rfp's"

Tguntenaar:
Ori and Trentin pointed out that there are some issues so I will revert the changes until this is investigated better. 
Reverts NEAR-DevHub/neardevhub-bos#982


Tguntenaar:
There is one test failling, we know that has to do with the RPC. We should start using the archival RPC since the tx being fetched is older than 9 months

PR #1025, opened 2025-02-25
Upgrade playwright to fix tests

Megha-Dev-19:
Updated ubuntu to use `22.04`, since playwright fails to get installed on later versions as mentioned [here](https://github.com/microsoft/playwright/issues/30368#issuecomment-2669259798). I tried upgrading playwright but the same error.


Tguntenaar:
I will close this PR. Upgraded to playwright 1.50.1 in the PR: #1028

